PR #26, opened 2018-07-27
fix file path completion

fannheyward:
What I want: when entered `/` file trigger, items will popup for completion.

This maybe a **quick-but-dirty** fix, I just found that if `synName` is empty, `shouldComplete` will return false to disable completion. So I just add this check fix.

BTW, what's the effects of `synName string|comment` check?


chemzqm:
It's used for avoid popup menu shown when using `//`for comment.

PR #57, opened 2018-08-23
fix typescript.jsx languageModeId

QuadeMorrison:
In [b3604](https://github.com/neoclide/coc.nvim/commit/b3604e00d133e29777591349d352ae22b2c0015c#diff-375d26229ede96ca813dae07d7bdcf8a) prior to adding the typescriptreact type, it referenced 'typescript.jsx' not 'typescript.tsx'. I'm unsure of whether you intended for this, but it broke functionality for me.

If you'd prefer me to go through and rename all the variables to typescriptjsx or even add a new variable all together in all the necessary areas, I can do that for you, just let me know.

### DIFF
--- a/src/extensions/tsserver/features/bufferSyncSupport.ts
+++ b/src/extensions/tsserver/features/bufferSyncSupport.ts
@@ -20,6 +20,8 @@ function mode2ScriptKind(
       return 'TS'
     case languageModeIds.typescripttsx:
       return 'TSX'
+    case languageModeIds.typescriptjsx:
+      return 'TSX'
     case languageModeIds.typescriptreact:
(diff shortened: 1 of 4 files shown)


chemzqm:
The rename would break for user who're using https://github.com/ianks/vim-tsx/blob/master/ftdetect/typescript.vim as syntax plugin for jsx suppport of typescript, it could be added, just like `typescriptreact`

QuadeMorrison:
I've updated the code to include another variable rather than replacing the old one. Just for reference this is so users wont have issues if they use the follow plugin: https://github.com/peitalin/vim-jsx-typescript.

Issue #91, opened 2018-09-24
[Feature Request] Provide distinct highlight group-names for each diagnostic severity


gbcreation:
**Is your feature request related to a problem? Please describe.**
There is only one highlight group-name (`CocUnderline`) which is used for all types of diagnostic regardless of their severity. This does not allow to use different type of highlighting on screen to indicate their severity.

**Describe the solution you'd like**
It could be solved by replacing the `CocUnderline` highlight group by the following ones: `CocDiagnosticError`, `CocDiagnosticWarning`, `CocDiagnosticInfo` and `CocDiagnosticHint`. Each one would be associated to the corresponding diagnostic severity.

PR #92, opened 2018-09-24
Distinct highlight groups per severity

gbcreation:
This is a tentative to implement #91 

New highlight groups are `CocErrorHighlight`, `CocWarningHighlight`, `CocInfoHighlight` and `CocHintHighlight`.

The `CocUnderline` highlight group is removed.

Successfully tested on Neovim 0.3.1

### DIFF
--- a/plugin/coc.vim
+++ b/plugin/coc.vim
@@ -129,11 +129,14 @@ function! s:Enable()
   command! -nargs=0 CocErrors  :call coc#rpc#show_error()
 endfunction
 
-hi default CocErrorSign   guifg=#ff0000
-hi default CocWarningSign guifg=#ff922b
-hi default CocInfoSign    guifg=#fab005
-hi default CocHintSign    guifg=#15aabf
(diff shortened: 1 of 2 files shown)


plugin/coc.vim
```diff
@@ -129,11 +129,14 @@ function! s:Enable()
...
-hi default CocHintSign    guifg=#15aabf
-hi default CocUnderline   term=underline gui=underline
+hi default CocErrorSign        guifg=#ff0000
+hi default CocWarningSign      guifg=#ff922b
+hi default CocInfoSign         guifg=#fab005
+hi default CocHintSign         guifg=#15aabf
+hi default CocErrorHighlight   term=underline gui=underline
+hi default CocWarningHighlight term=underline gui=underline
+hi default CocInfoHighlight    term=underline gui=underline
+hi default CocHintHighlight    term=underline gui=underline
```
chemzqm: I would prefer to use `hi link` like:
```
hi default link CocErrorSign CocUnderline
```
so it doesn't break.

PR #203, opened 2018-11-28
Clear diagnostic info when restarting

daquexian:
After restarting, the diagnostic info is not cleared, so that airline (and maybe other status lines) still shows the original error and warning number. However, the errors and warnings may be gone because of the new configuration of the language server. In these cases, the error and warning number in the status line will not be cleared until new errors or warnings appear.

### DIFF
--- a/autoload/coc/rpc.vim
+++ b/autoload/coc/rpc.vim
@@ -156,6 +156,7 @@ endfunction
 
 function! coc#rpc#restart()
   call coc#util#clear_signs()
+  call coc#util#clear_diagnostic_info()
   if has('nvim')
     let [code] = jobwait([s:channel_id], 100)
     " running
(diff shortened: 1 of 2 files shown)


daquexian:
Also, I found that even though `clear_signs()` is called when restarting, it only list all _visible_ buffers in `buflist`. So I fixed it.

However, I have no idea about how to clear diagnostic info of all buffers. I'm new to vim plugin development :)

What's more, the diagnostic highlights (Sorry I don't know the formal term) of both visible buffers and invisible buffers are not cleared. Could you please give some tips about how to clear them?

chemzqm:
Thanks for reporting, the clear not works because the buffer is cleared from workspace before diagnostic refresh. I will fix this problem.

PR #207, opened 2018-11-28
Fix too many args to rpc_installed

brianembry:
Was seeing this trying to use yarn to install vim-node-rpc.

```
TERMINAL  yarn global add vim-node-rpc                                                                                                                                                                                                                                                           ☰    1  /4 ㏑
Error detected while processing function <SNR>88_OnExit:
line    6:
E118: Too many arguments for function: <SNR>88_rpc_installed
```


chemzqm:
Thanks.

PR #316, opened 2019-01-03
feat(configuration): add a hook where coc#config can be called

free-easy:


chemzqm:
But this hook is not needed, function `coc#config` could be called at any time:
```
function! coc#config(section, value)
  let g:coc_user_config[a:section] = a:value
  call coc#rpc#notify('updateConfig', [a:section, a:value])
endfunction
```

free-easy:
Hi,
I forgot to specify that I called `coc#config` with following parameters

```vim
coc#config("languageserver", {
    'ccls': {
         'args': ['--log-file=/tmp/ccls.log', '--init="{\"cacheDirectory\": \"/tmp/ccls\"}"'],
         'filetypes': ['c', 'cpp'],
         'command': 'ccls'
    },
     'typescript-language-server': {
         'args': ['--stdio'],
         'filetypes': ['typescript'],
         'command': 'typescript-language-server'
     }
  }
)
```
This seems to work only if `coc#config` is called before the service is initialized.

chemzqm:
* languageserver is not watched for change event, it could be improved.

* Why you need to add languageserver configuration after VimEnter?

* CocSetup doesn't help when the server already started, since it's fired before you apply the configuration.

* If you're using lazyload, you should use the hook that provided by the loader plugin.

free-easy:
Thanks for implementing `languageserver` watch configuration!

free-easy:
f

Issue #353, opened 2019-01-18
coc#add_extension() can cause "Too Many arguments" error


docwhat:
**Result from CocInfo**
Run `:CocInfo` command and paste the content below.

**Describe the bug**
Given this vim script:

```viml
let s:coc_extensions = [
      \  'coc-css',
      \  'coc-emmet',
      \  'coc-highlight',
      \  'coc-html',
      \  'coc-dictionary',
      \  'coc-emoji',
      \  'coc-omni',
      \  'coc-tag',
      \  'coc-pairs',
      \  'coc-java',
      \  'coc-json',
      \  'coc-pyls',
      \  'coc-rls',
      \  'coc-snippets',
      \  'coc-solargraph',
      \  'coc-yaml',
      \  'coc-eslint',
      \  'coc-jest',
      \  'coc-prettier',
      \  'coc-tslint',
      \  'coc-tsserver',
      \ ]


call coc#add_extension(join(get(s:, 'coc_extensions', [])))
```

It causes these messages:

```
extension coc-css coc-emmet coc-highlight coc-html coc-dictionary coc-emoji coc-omni coc-tag coc-pairs coc-java coc-json coc-pyls coc-rls coc-snippets coc-solar
graph coc-yaml coc-eslint coc-jest coc-prettier coc-tslint coc-tsserver installed!
Error detected while processing function <lambda>6[1]..<SNR>77_OnExit[6]..<lambda>5[1]..<SNR>77_OnExtensionInstalled[3]..coc#rpc#notify[2]..<SNR>78_notify:
line    7:
E699: Too many arguments
Press ENTER or type command to continue
```

**To Reproduce**

1. Add the above vim script to your `.vimrc` to run after the plugins are installed.

chemzqm:
I don't know what's wrong with your vim, just tried to fix it.
Paste `:CocInfo` if the problem still exists.

docwhat:
```
## versions

vim version: NVIM v0.3.4
node version: v10.14.1
coc.nvim version: 0.0.46
term: tmux-256color
platform: darwin

## Error messages

```

After digging in a bit it looks like it blows up in `coc#rpc#notify(method, args)`.

It is called with `method = 'CocInstalled'` and the `args` are:

1. `coc-css`
1. `coc-emmet`
1. `coc-highlight`
1. `coc-html`
1. `coc-dictionary`
1. `coc-emoji`
1. `coc-omni`
1. `coc-tag`
1. `coc-pairs`
[...]

PR #356, opened 2019-01-19
fix(extensions): handle more than 19 extensions

docwhat:
If you use coc#add_extension() with more than 19 extensions, it can
cause coc#util#install_extension() to call coc#rpc#notify() with more
than 20 arguments, generating a `E699: Too many arguments` error.

fixes: #353


chemzqm:
Just send string to server should be fine, I will fix it.

Issue #360, opened 2019-01-19
Neovim's virtual text for diagnostic messages


kristijanhusak:
**Is your feature request related to a problem? Please describe.**
It's more nice to have, but a great one for me.
I would like to use Neovim's virtual text feature to show diagnostic messages.
ALE has a support for [this](https://github.com/w0rp/ale/blob/a50e049e747739c711198bd706c7340e16b182cc/doc/ale.txt#L1936) for some time now.

**Describe the solution you'd like**
This should be an option, the same way `codeLens` is.

**Describe alternatives you've considered**
Alternative is to install ALE, and pass down messages to it. I would like to avoid installing ALE only because of this, because `coc` covers everything that i need.

Avi-D-coder:
I had the same thought. Working on it now!

chemzqm:
Will add an option for that, but I have to say that redraw of virtual text could affect your type experience.

Avi-D-coder:
@chemzqm I Have it almost done. I just have to add the finishing touches.

kristijanhusak:
@chemzqm wouldn't triggering the virual text redraw on CursorHold fix the issue? Then it wouldn't redraw so often, at least not while typing.

Avi-D-coder:
@kristijanhusak that's a great idea. I will change my prototype.

chemzqm:
Another problem is how to deal with multiple lines diagnostic message?
What if there're several diagnostics at same line?

Avi-D-coder:
Multiline for now is out of the scope of my implementation.
When there are several diagnostics at same line one of the highest severity should be shown.

Avi-D-coder:
@kristijanhusak Any additional thoughts, or can the issue be closed?

kristijanhusak:
@Avi-D-coder i left a comment on the PR https://github.com/neoclide/coc.nvim/pull/365#issuecomment-455888062. Could we add something like this?

Avi-D-coder:
chemzqm added a one space prefix. I'll make a PR that allows you to set the prefix.

Avi-D-coder:
@kristijanhusak how does https://github.com/neoclide/coc.nvim/pull/375 look?

kristijanhusak:
Looks great, thanks!

PR #365, opened 2019-01-20
Implement NeoVim virtual text diagnostics

Avi-D-coder:
solves: https://github.com/neoclide/coc.nvim/issues/360

Uses NeoVim virtual text to display diagnostics.
`"coc.preferences.diagnostic.virtualText"` defaults to `false`.

### DIFF
--- a/src/diagnostic/buffer.ts
+++ b/src/diagnostic/buffer.ts
@@ -58,6 +58,7 @@ export class DiagnosticBuffer {
     })
     this.nvim.pauseNotification()
     sequence.addFunction(this.setDiagnosticInfo.bind(this, diagnostics))
+    sequence.addFunction(this.addVTextDiagnostics.bind(this, diagnostics))
     sequence.addFunction(this.setLocationlist.bind(this, diagnostics))
     sequence.addFunction(this.addSigns.bind(this, diagnostics))
     sequence.addFunction(this.addHighlight.bind(this, diagnostics))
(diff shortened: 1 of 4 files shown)


src/diagnostic/buffer.ts
```diff
@@ -1,5 +1,5 @@
 import { NeovimClient as Neovim } from '@chemzqm/neovim'
-import { Diagnostic, DiagnosticSeverity, Range } from 'vscode-languageserver-protocol'
+import { Diagnostic, DiagnosticSeverity, Range, ApplyWorkspaceEditRequest } from 'vscode-languageserver-protocol'
```
chemzqm: `ApplyWorkspaceEditRequest ` is not used.

src/diagnostic/buffer.ts
```diff
@@ -172,6 +174,39 @@ export class DiagnosticBuffer {
...
   }
 
+  // Add a virtual text diagnostic if in neovim
+  private async addVTextDiagnostics(diagnostics: Diagnostic[]) {
+    if (this.isVim || !this.config.virtualText) {
+      return
+    }
+
+    let { bufnr, nvim, vTextNameSpace } = this
+    let newVTextNameSpace = await workspace.createNameSpace();
```
chemzqm: Namespace only need to be created once.
Could be created from diagnostic manager and pass to diagnostic buffer.

src/diagnostic/buffer.ts
```diff
@@ -172,6 +174,39 @@ export class DiagnosticBuffer {
...
+
+    for (let diagnostic of diagnostics) {
+      let setVirtualText = (highlight:String) => {
+          nvim.call("nvim_buf_set_virtual_text", [bufnr, newVTextNameSpace, diagnostic.range.start.line, [[diagnostic.message, highlight]], {}])
```
chemzqm: Should not use `nvim.call` here, each call would cause neovim redraw.
Add `false` as third argument to use notification or use `buffer.setVirtualText` which use notification.

kristijanhusak:
@Avi-D-coder This works great. Can we add just one more option that will add some prefix to the message, or some padding? It looks too close like this:

screenshot

This option could maybe reflect any virtual text usage (codeLens and diagnostic). That decision is up to @chemzqm .

PR #381, opened 2019-01-27
Change example goto type mapping to gt

Avi-D-coder:
Currently the example `vimrc` maps goto type to `gy`, but `gt` seams like a more intuitive mapping.

Thoughts?


chemzqm:
`gt` is builtin for switch tab page, so I don't recommend to overwrite it.

Avi-D-coder:
Good point. I forgot `:map` does not work with builtins.

Issue #373, opened 2019-01-23
Incorrect cursor placement after completion


Avi-D-coder:
## versions

vim version: NVIM v0.3.4
node version: v11.7.0
coc.nvim version: 0.0.46
term: xterm-256color
platform: linux
built from list branch

**Describe the bug**
Completions with an additional prefix cause the cursor to end up short, by the difference between text shown in completion menu and text actually inserted.
![bad-cursor](https://user-images.githubusercontent.com/29133776/51620245-f6f11800-1eff-11e9-8a4c-e1fb68bd1c98.gif)

Avi-D-coder:
I'm looking into fixing this

chemzqm:
Can't reproduce it.

Avi-D-coder:
I wiped all state in the project and `rm -rf ~/.config/nvim/plugged/coc.nvim/`. Before building from scratch and the issue persists.

Does the nvim completion item or coc place the cursor after completion is triggered?

PR #384, opened 2019-01-28
fix(completion): correct cursor placement after additional edits

Avi-D-coder:
https://github.com/neoclide/coc.nvim/issues/373


chemzqm:
Will add some test for it.

PR #391, opened 2019-01-30
Config option to display multiple lines of a diagnostic in a single virtual text line

Avi-D-coder:
`coc.preferences.diagnostic.virtualTextLines`  allows the user to specify the number of non empty lines from each diagnostic message that should be shown in a single virtual text line.
`coc.preferences.diagnostic.virtualTextLineSeparator` allows the users to specify a custom `string` to go in between the conjoined lines.

This commit does not change any defaults, however I would advocate `virtualTextLines` defaulting to three instead of one. This would allow the standard `mismatched type\n {expected type}\n {found type}` diagnostics to be correctly displayed.

### DIFF
--- a/src/diagnostic/buffer.ts
+++ b/src/diagnostic/buffer.ts
@@ -188,11 +188,11 @@ export class DiagnosticBuffer {
       let highlight = getNameFromSeverity(diagnostic.severity) + 'Sign'
       let msgLines = diagnostic.message.split(/\n/)
       let msg = msgLines[0]
-      for (let i = 1; i < Math.min(this.config.virtualTextLines, msgLines.length); i++) {
-        if (msgLines[i] != '') {
-          msg += this.config.virtualTextLineSeparator + msgLines[i].trim()
-        }
-      }
+      msgLines
+        .map((l: string) => l.trim())
+        .filter((l: string) => l !== '')
+        .slice(1, this.config.virtualTextLines)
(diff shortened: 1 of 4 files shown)


src/diagnostic/buffer.ts
```diff
@@ -186,7 +186,13 @@ export class DiagnosticBuffer {
...
       lines.add(line)
       let highlight = getNameFromSeverity(diagnostic.severity) + 'Sign'
-      let msg = diagnostic.message.split(/\n/)[0]
+      let msgLines = diagnostic.message.split(/\n/)
+      let msg = msgLines[0]
+      msgLines
+        .map((l: string) => l.trim())
+        .filter((l: string) => l !== '')
+        .slice(1, this.config.virtualTextLines)
+        .forEach((l: string) => msg += this.config.virtualTextLineSeparator + l)
```
chemzqm: Should be like this:
``` js
let msg = msgLiens.map(l => l.trim()).filter(l => l.length > 0).slice(0, this.config.virtualTextLines).join(this.config.virtualTextLineSeparator)
```

chemzqm:
> virtualTextLines defaulting to three instead of one.

Agree, I think 3 is better.

src/diagnostic/buffer.ts
Avi-D-coder: I had forgotten javascript has a `intersperse` equivalent. Thanks. I fixed it now.

Avi-D-coder:
@chemzqm  does this need anything else?

PR #410, opened 2019-02-07
Add keepCompleteopt configuration option. Closes #406.

### LINKED ISSUE #406: Add option to auto insert first item when doing manual trigger
I want to use manual trigger for showing a completion (`coc.preferences.autoTrigger: none`), and i tried using `coc.preferences.noselect: false`, but it does only half of job for me.

I would like it to behave like vim's native completion when you press `<c-n>`, with `set completeopt=menu`. Basically to automatically select and insert the first item. `coc.preferences.noselect` only selects the first item, without inserting it.

Nothing.

**Additional context**
This issue is related to this one https://github.com/neoclide/coc.nvim/issues/124, but it's not a duplicate because that one is solved with `coc.preferences.noselect`.

Also, i understand this option would mess up auto triggered completion, so i guess this option should work and be limited to manual triggering.

kristijanhusak:


src/completion/index.ts
```diff
@@ -490,9 +492,9 @@ export class Completion implements Disposable {
   }
 
   private get completeOpt(): string {
-    let { noselect } = this.config
+    let { noselect, noinsert } = this.config
     let preview = workspace.completeOpt.indexOf('preview') !== -1
-    return `${noselect ? 'noselect,' : ''}noinsert,menuone${preview ? ',preview' : ''}`
+    return `${noselect ? 'noselect,' : ''}${noinsert ? 'noinsert' : ''},menuone${preview ? ',preview' : ''}`
```
chemzqm: What if user want to use `menu` instead of `menuone`?

chemzqm:
I would prefer a setting named: `coc.preference.keepCompleteopt`, when it's true not change user's completeopt, and when `completeopt` doesn't have `noinsert`, force auto trigger to `none`.

kristijanhusak:
@chemzqm should then also `noselect` be removed in favor of `keepCompleteopt`, or just to be added and respected accordingly?

chemzqm:
I prefer just keep `noselect` and not make it work when `keepCompleteopt` is true.

kristijanhusak:
@chemzqm i added `keepCompleteopt`.

Issue #435, opened 2019-02-13
Possible to not use cmdheight=2 & wrap message text & ale?


tomspeak:
Hi,

1. I was wondering if it's possible to not have to set `cmdheight=2` without having to deal with the message `Press enter to continue`?

2. Is it possible to have the content put into the command status wrap? It currently cuts off. Flow seems to have very long messages.

3. Can you explain how this works in conjunction with Ale? Some of the functionality is blurred between the two plugins. I have seen you use Ale (if that's still true), do you have any advice on how to get the best out of each plugins strengths?

Thank you!

chemzqm:
1. `cmdheight=1` is not required, let me know how to trigger `1. Press enter to continue` when it's 1.

2. What content do you mean?

3. I've disabled ale from lint the files that coc could handle, but you can disable coc diagnostic if your language server is slow, or make ale display diagnostics from coc by settings.

tomspeak:
1. Odd! `cmdheight=1` being set seems to fix the issue? It would only work with `cmdheight=2` last night. Now the error message `Press enter to continue` does not appear.

2. Sorry, that was not very clear. For long text (mostly Flow errors), the messages are cut off, not wrapped.
![screenshot 2019-02-14 at 10 44 37 am](https://user-images.githubusercontent.com/2415569/52782329-b4c27e80-3046-11e9-9597-66d6ae84b974.png)

3. Thank you!

chemzqm:
> 2\. the messages are cut off, not wrapped.

It would be truncated on CursorHold, but the picture you give shows full message.
Checkout `:h coc-action-diagnosticInfo` or use `<Plug>(coc-diagnostic-next)` and `<Plug>(coc-diagnostic-prev)` when you need diagnostic messages without truncate.

tomspeak:
Sorry, I should have shown a message being cut off. 

Your fix works. Thanks for your help.

tomspeak:
![screenshot 2019-02-14 at 3 02 49 pm](https://user-images.githubusercontent.com/2415569/52795450-b81b3180-3069-11e9-8674-ff9381f1e044.png)

I seem to have spoken too soon.

When hovering over a warning with `cmdheight=1`, this prompt appears and forces you to press "Enter" — is this possible to get around without increasing `cmdheight`?

Lastly, you can see the text being cut off, I assume there's no way for this to automatically wrap with `cmdheight=1`?

chemzqm:
[...]

PR #492, opened 2019-03-07
fix(diagnostics): echo max length

pyrho:
On neovim the maximum length of an echo'd line seems to be `columns -
12` (where columns is the vim option), passed that length it will trigger the "Press ENTER or type a
command" prompt.

On vim this value seems to be `columns - 8`.

This fixes https://github.com/neoclide/coc.nvim/issues/435 .

### DIFF
--- a/src/workspace.ts
+++ b/src/workspace.ts
@@ -534,10 +534,16 @@ export class Workspace implements IWorkspace {
       let last = lines[cmdHeight - 1]
       lines[cmdHeight - 1] = `${last} ...`
     }
-    let columns = await nvim.getOption('columns')
+
+    // If an echoed line exceeds `columns - 9` for vim or `columns - 12` for nvim
+    // The "Press ENTER or type command to continue" prompt will trigger.
+    // This may be a bug in vim/nvim
+    // These numbers were found empirically.
+    let maxLength = await nvim.getOption('columns').then((c: number) => c - (this.isVim ? 9 : 12))
+
     lines = lines.map(line => {
       line = line.replace(/\n/g, ' ')
-      if (truncate) line = line.slice(0, (columns as number) - 1)
+      if (truncate) line = line.slice(0, maxLength)
       return line
     })
     nvim.callTimer('coc#util#echo_lines', [lines], true)


pyrho:
The following example code triggers the issue:
```
interface Browser {
    aName1: (args: { aLongName1: string; aLongName2: number, aLongName3: string, aLongName4: string, aLongName5: number, aLongName6: string, aLongName7: string }) => Promise<void>;
}

const b: Browser = {
    aName1: () => {
        return 1;
    }
}
```
Placing the cursor on `aName1` and holding will trigger the echo'ing of the diagnostic, triggered the "Press ENTER" prompt.

pyrho:
I've found the "magic" numbers empirically by running both vim and nvim with `-u NONE` and trying to echo the longest string of characters without triggering the "Press ENTER" prompt.

I don't understand the behavior, it may be a bug within vim itself, but I'm not familiar with that codebase.

fannheyward:
`set cmdheight=2` in vimrc/init.vim

pyrho:
`cmdheight` does not help, see issue https://github.com/neoclide/coc.nvim/issues/435 .

Also if you look at the code (from master):
```
    if (lines.length > cmdHeight && truncate) {
      lines = lines.slice(0, cmdHeight)
      let last = lines[cmdHeight - 1]
      lines[cmdHeight - 1] = `${last} ...`
    }
    let columns = await nvim.getOption('columns')
    lines = lines.map(line => {
      line = line.replace(/\n/g, ' ')
      if (truncate) line = line.slice(0, (columns as number) - 1)
      return line
    })
```

You will see that `cmdheight` value only affects the number of lines displayed, not the line length.

pyrho:
This updated example shows the issue, even with `cmdheigth=2`:
```
interface Browser {
    aName1: (args: { aLongName1: string; aLongName2: number, aLongName3: string, aLongName4: string, aLongName5: number, aLongName6: string, aLongName7: string, aLongName8: string, aLongName9: string }) => Promise<void>;
}

interface RetType {
    aLongName1: string; aLongName2: number, aLongName3: string, aLongName4: string, aLongName5: number, aLongName6: string, aLongName7: string, aLongName8: string, aLongName: string;
}

const b: Browser = {
    aName1: () => {
        return {} as RetType;
    }

}
```

Make sure you make your vim window small for this issue to be witnessed.
In my case, with the above code `set columns?` returns 97.

screenshot 2019-03-07 at 13 51 35

chemzqm:
Thanks, I've fixed this issue on master.
If you want to better experience, use master of neovim for floating support.

pyrho:
Awesome thanks, for future reference here is the changeset linked to this: https://github.com/neoclide/coc.nvim/commit/e73474c32726716a34d5678ee0902dd6bee31c35.

Side note, but since these numbers (`9` and `12`) really feel like they're coming out of nowhere in the code, a comment might help future you and future contributors to remember what this is about.

PR #543, opened 2019-03-17
fix(floating): Fix floating due to neovim api changes

kozer:
Due to neovim api changes call to ```nvim.openFloatWindow``` should change accordingly


chemzqm:
It can't compile, upgrade of neovim client needed.

Issue #600, opened 2019-03-26
Map custom glyphs to item kinds


gbrlsnchs:
**Is your feature request related to a problem? Please describe.**
It would be cool (but not necessary) to have icons shown inside the pum (or wherever they appear) instead of single chars like `v`, `f`, etc.

**Describe the solution you'd like**
Use [Vim DevIcons](https://github.com/ryanoasis/vim-devicons) to, when available, substitute the letters. Custom icon remapping would also be nice to have.

**Additional context**
I'm ignorant when it comes to the LSP and how this plugin maps types to each option, so maybe this is not even possible at all. Sorry if that's the case.

chemzqm:
The problem is DevICons only have icons for filetypes, not for complete item kind.

gbrlsnchs:
You're right, sorry. What about mapping kinds directly to a custom string, this way one is able to set whatever glyph or text they want. No DevIcons needed at all.

gbrlsnchs:
@chemzqm I am playing around with this a little:
![image](https://user-images.githubusercontent.com/7577128/55033045-ff91c780-4ff0-11e9-99ea-c10823bc0b0c.png)
![image](https://user-images.githubusercontent.com/7577128/55033081-159f8800-4ff1-11e9-88de-6aaaed818cd1.png)

I think it's just a matter of enabling this in the `coc-settings.json` file and, if not explicitly mapped there, fallback to default values. What do you think?

chemzqm:
Looks great, consider send a PR.

gbrlsnchs:
@chemzqm What's the proper way to retrieve configuration values? I need to get it from inside `src/util/complete.ts`, but I could not grasp how to do so yet.

chemzqm:
You can set the configuration at https://github.com/neoclide/coc.nvim/blob/master/src/completion/index.ts#L147 and use it at https://github.com/neoclide/coc.nvim/blob/master/src/completion/complete.ts#L193

gbrlsnchs:
Should the custom config be under `"suggest"` or `"coc.preferences"`?

chemzqm:
under `"suggest" `

gbrlsnchs:
> https://github.com/neoclide/coc.nvim/blob/master/src/completion/complete.ts#L193

This seems to set an already resolved kind. Isn't it better to retrieve the custom config once in https://github.com/neoclide/coc.nvim/blob/master/src/util/complete.ts by setting
```typescript
[...]

PR #603, opened 2019-03-27
Allow kind labels to be configured via coc-settings.json

gbrlsnchs:
This implements what's suggested [here](https://github.com/neoclide/coc.nvim/issues/600), that is to enable a user to configure custom labels for kinds that appear in the completion window.

-- EDIT --
This makes the popup menu a little more good-looking:
image
image
image
image

Here's a simple configuration for the pictures above:
```json
{
  "suggest.completionItemKindLabels": {
    "function": "\uf794",
    "method": "\uf6a6",
    "variable": "\uf71b",
    "constant": "\uf8ff",
    "struct": "\ufb44",
    "class": "\uf0e8",
    "interface": "\ufa52",
    "text": "\ue612",
    "enum": "\uf435",
    "enumMember": "\uf02b",
    "module": "\uf668",
    "color": "\ue22b",
    "property": "\ufab6",
    "field": "\uf93d",
    "unit": "\uf475",
    "file": "\uf471",
    "value": "\uf8a3",
    "event": "\ufacd",
    "folder": "\uf115",
    "keyword": "\uf893",
    "snippet": "\uf64d",
    "operator": "\uf915",
    "reference": "\uf87a",
    "typeParameter": "\uf278",
    "default": "\uf29c"
  }
}
```


tomspeak:
Thank you for your work on this. Looks great.

yatli:
@gbrlsnchs may I ask what font patches do you use? I tried latest nerdfont but it's missing some symbols. (It's also possible that my ui implementation is not picking up the correct code plane)

gbrlsnchs:
> @gbrlsnchs may I ask what font patches do you use? I tried latest nerdfont but it's missing some symbols. (It's also possible that my ui implementation is not picking up the correct code plane)

I use the Nerd Font one, in my machine it's named `SauceCodePro Nerd Font Mono`. It has every symbol listed [here](https://nerdfonts.com/#cheat-sheet).

yatli:
I was using an incomplete set. parsing the full planes now.
Thanks!!

fetchusername:
This is awesome! thank you for adding this.

I would like to get the unicode character for `variable`, `field` as shown [here](https://code.visualstudio.com/docs/editor/intellisense)

How do I find that?

daquinoarthur:
Thanks a lot for your work on this!

PR #623, opened 2019-03-30
Turn on job control when installing

nhynes:
If `-m` isn't passed, shells which turn job control off when non-interactive (e.g. dash, the default sh on ubuntu) will error when installing coc. This PR fixes this issue by always turning on job control (via `-m`).


chemzqm:
Thanks.

PR #647, opened 2019-04-03
Some improvments

mihaelartz:


chemzqm:
Thanks, but it's not needed.

Issue #650, opened 2019-04-04
coc-settings: expand $HOME or ~ for language server executable paths


panders23:
Hi, 

As far as I tried its not possible to use ~ or $HOME in coc-settings.json when registering custom language server's. I stumbled on https://github.com/neoclide/coc.nvim/wiki/Using-configuration-file#default-coc-preferences
[quote] there're also some vim global variables used for configuration, check out :h coc-variable[/quote] but I don't know how to access it.

Currently my workaround is to add every LS executable to $PATH. 

Support for this would be nice as i share coc-settings between 2 machines and i think this would make dotfile sharing generally easier.

chemzqm:
Checkout `:h coc#config()` if you want use vim's funciton/variable with coc configuration.

panders23:
@chemzqm is it possible to expand  ~ for args also?
I fiddled with it locally and so far this works but I'm no typescript expert:
```
diff --git a/src/language-client/index.ts b/src/language-client/index.ts
index f9022b8..400253e 100644
--- a/src/language-client/index.ts
+++ b/src/language-client/index.ts
@@ -410,13 +410,20 @@ export class LanguageClient extends BaseLanguageClient {
       }
     } else if (Executable.is(json) && json.command) {
       let command: Executable = json as Executable
-      let args = command.args || []
+      let command_args_ro = command.args || []
+      let args = []
+
       let options = Object.assign({}, command.options)
       options.env = options.env ? Object.assign(options.env, process.env) : process.env
       options.cwd = options.cwd || serverWorkingDir
       if (command.command.startsWith('~')) {
         command.command = command.command.replace(/^~/, os.homedir())
       }
+
+      for (var i = 0, len = command_args_ro.length; i < len; i++) {
+          args[i] = command_args_ro[i].replace(/^~/, os.homedir())
+      }
+
       let serverProcess = cp.spawn(command.command, args, options)
       if (!serverProcess || !serverProcess.pid) {
         throw new Error(`Launching server using command ${command.command} failed.`)
```

unphased:
Hi, this works, but not recursively. See: 

```
  , "clangd.path": "~/.config/coc/extensions/coc-clangd-data/install/13.0.0/clangd_13.0.0/bin/clangd"
  , "eslint.options": { "overrideConfigFile": "~/.eslintrc.json"  }
```
Clangd is happy, but eslint's config doesn't get picked up and complains of 
```
[...]

PR #678, opened 2019-04-08
expand tilde in languageserver args #650

panders23:
Continuation of #650 

Also expand ~ for language server language-server args:

for example:
```json
"languageserver": {
        "phppsalmls": {
            "command": "php",
            "args": ["~/dev/php/composer/vendor/vimeo/psalm/psalm-language-server"],
            "filetypes": ["php"]
        }
}
```


chemzqm:
Add `"shell": true`  to configuration if you want to expand arguments.

Issue #512, opened 2019-03-12
Function/method name in statusline


robertocarta:
Hi - i'm using this https://github.com/tyru/current-func-info.vim to get the current function/method name in my statusline. Would it be possible to achieve something similar with coc?

chemzqm:
Sure, you can just query for document symbols and find the nearest function symbol, but it doesn't have params information, need to extract from the line contains that function.

liuchengxu:
hello, @robertocarta , I have implemented this feature in [vista.vim](https://github.com/liuchengxu/vista.vim). Here is the PR waiting to be merged https://github.com/liuchengxu/vista.vim/pull/28 .

<img width="1680" alt="屏幕快照 2019-03-23 下午8 03 25" src="https://user-images.githubusercontent.com/8850248/54865892-efd37280-4da7-11e9-9970-6e761c798d99.png">

You can show the nearest method/function in the statusline as follows:

```vim
function! NearestMethodOrFunction() abort
  return exists('b:vista_nearest_method_or_function') ? b:vista_nearest_method_or_function : ''
endfunction

set statusline+=%{NearestMethodOrFunction()}
```

Also refer to https://github.com/liuchengxu/eleline.vim/pull/18.

andersevenrud:
In case anyone uses lightline, this is how you can use the above (modified example of README in this repo):

```vim
let g:lightline = {
      \ 'active': {
      \   'left': [ [ 'mode', 'paste' ],
      \             [ 'cocstatus', 'readonly', 'filename', 'modified', 'vistanearest' ] ]
      \ },
      \ 'component_function': {
      \   'cocstatus': 'coc#status',
      \   'vistanearest': 'NearestMethodOrFunction'
      \ }
 \}
```

@liuchengxu Is there any way to "silently" trigger vista.vim ? Because it seems like I have to open and close it in order for this to work. If I haven't missed something really obvious that is....

liuchengxu:
@robertocarta No way at present. That feature is also what I want, you can file an issue in vista.vim and I'll work on it when I get some time.

andersevenrud:
@liuchengxu Thanks. I'll take the discussion to your repo.

pyrho:
Inspired by @liuchengxu 's work on Vista I started to look into coc's code to see how to implement the feature natively in coc.

[...]

PR #679, opened 2019-04-08
feat(plugin): current function symbol

pyrho:
tl;dr
====
* New plugin action: `getCurrentFunctionSymbol`
* New preference: `coc.preferences.currentFunctionSymbolAutoUpdate`
* New depedency on `binary-search`.

---

A new handler method had been added (`getCurrentFunctionSymbol`) that
returns the name of the enclosing
Class/Method/Function/Interface/Enum/Constructor relative to the current
cursor position.
It will also set the `b:coc_current_function` buffer variable.

If the user choses to, by setting the
`coc.preferences.currentFunctionSymbolAutoUpdate` to `true` (default:
false), a hook is setup on the `CursorHold` event that will call
`getCurrentFunctionSymbol` and subsequently update the value of
`b:coc_current_function`.
This is useful in the context of an enhanced status line for instance.

Fixes https://github.com/neoclide/coc.nvim/issues/512

Inspired by @liuchengxu 's work on https://github.com/liuchengxu/vista.vim .

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -1,4 +1,5 @@
 import { NeovimClient as Neovim } from '@chemzqm/neovim'
+import binarySearch from 'binary-search'
 import { CancellationTokenSource, CodeAction, CodeActionContext, CodeActionKind, Definition, Disposable, DocumentHighlight, DocumentHighlightKind, DocumentLink, DocumentSymbol, ExecuteCommandParams, ExecuteCommandRequest, Hover, Location, LocationLink, MarkedString, MarkupContent, Position, Range, SymbolInformation, SymbolKind, TextEdit, SelectionRange } from 'vscode-languageserver-protocol'
 import Uri from 'vscode-uri'
 import commandManager from '../commands'
@@ -62,6 +63,7 @@ interface Preferences {
   hoverTarget: string
   previewAutoClose: boolean
   bracketEnterImprove: boolean
+  currentFunctionSymbolAutoUpdate: boolean
 }
 
 export default class Handler {
@@ -177,6 +179,11 @@ export default class Handler {
         nvim.command('pclose', true)
       }
     }, 100), null, this.disposables)
+    if (this.preferences.currentFunctionSymbolAutoUpdate) {
+      events.on('CursorHold', async () => {
+        await this.getCurrentFunctionSymbol()
+      }, null, this.disposables)
+    }
 
     let provider: TextDocumentContentProvider = {
       onDidChange: null,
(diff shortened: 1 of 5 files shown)


pyrho:
Example lightline integration:
```
function! CocFunc()
    let currentFunctionSymbol = get(b:, 'coc_current_function', '')
    return currentFunctionSymbol !=# '' ? "\uf6a6 " .currentFunctionSymbol : ''
endfunction

let g:lightline.component_function = {
            \ 'cocstatus': 'coc#status',
            \ 'currentFunc': 'CocFunc'
            \}
let g:lightline.active = {
            \ 'left': [ [ 'mode', 'paste', ],
            \           [ 'readonly', 'relativepath', 'modified' ],
            \           [ 'cocstatus', 'currentFunc' ] ],
           ....
           \}
```

Screenshot 2019-04-08 at 23 00 25

pyrho:
I didn't add any test because I was not sure on how to proceed, how would I test this (if doable) ?

src/handler/index.ts
```diff
@@ -195,6 +202,47 @@ export default class Handler {
     this.colors = new Colors(nvim)
   }
 
+  public async getCurrentFunctionSymbol(): Promise<string> {
+    let { position } = await workspace.getCurrentState()
+    const buffer = await this.nvim.buffer
+    let symbols = await this.getDocumentSymbols()
+
+    if (!symbols || symbols.length === 0) {
+        buffer.setVar('coc_current_function', '')
```
chemzqm: Should be `buffer.setVar('coc_current_function', '', true)`

src/handler/index.ts
```diff
@@ -195,6 +202,47 @@ export default class Handler {
...
+
+    const currentFunctionName = (() => {
+      const sym = symbols[symbolPosition]
+      if (!sym || (position.line > sym.range.end.line)) {
+        return ''
+      } else {
+          return sym.text
+      }
+    })()
+    buffer.setVar('coc_current_function', currentFunctionName)
```
chemzqm: Should be `buffer.setVar('coc_current_function', currentFunctionName, true)`

src/handler/index.ts
pyrho: Done.

src/handler/index.ts
pyrho: Done.

src/handler/index.ts
```diff
@@ -177,6 +179,11 @@ export default class Handler {
         nvim.command('pclose', true)
       }
     }, 100), null, this.disposables)
+    if (this.preferences.currentFunctionSymbolAutoUpdate) {
+      events.on('CursorHold', () => {
+        this.getCurrentFunctionSymbol()
```
chemzqm: Need change to `async  () => { await this.getCurrentFunctionSymbol()`, it will catch possible errors.

pyrho: ✔️

Issue #738, opened 2019-04-29
Don't use npx in npm scripts


Jomik:
i would prefer if you added `npm-run-all` to `devDependencies`, rather than using `npx` in an npm script.

This makes it quite a bit easier to resolve dependencies.

chemzqm:
You have options:

1. Use `:call coc#util#install()` to avoid build javascript which is recommended.
2. Use `npm install -g npm-run-all` to avoid resolve for npm-run-all.

PR #739, opened 2019-04-30
fix(yarn): add npm-run-all as devDependency

Jomik:
In relation to #738 
This feels like a better way to add dependencies that you need for building your plugin.
We still get the ability to run `npx npm-run-all clean build` from command line.
`yarn prepare` works as expected.
The dependency is clearly listed in `package.json`.


Jomik:
Thank you :)

PR #769, opened 2019-05-08
update install.cmd

yatli:
powershell module PSCX overrides Expand-Archive but does not have -DestinationPath.


chemzqm:
The problem is this script can't work when PSCX is not installed.

yatli:
Microsoft.PowerShell.Archive is a built in module and it provides Expand-Archive.

PR #784, opened 2019-05-18
Hide alert of changing 'completeopt' when it is not replaced.

miyanokomiya:
## Problem
I use "coc.nvim" and "[deoplete.nvim](https://github.com/Shougo/deoplete.nvim)".
"deoplete.nvim" often changes "completeopt" to "noselect,menuone" and "coc.nvim" show the alert message as the same time.
Then, the message is displayed at the command line for a moment as below.
May-18-2019 22-27-58

However, the value "noselect,menuone" is the same as that "coc.nvim" uses. So, the message is unnecessary.

## Proposal
Skip to show the message if new value equals old one.


chemzqm:
It's feature, it's not designed to work with other completion tool at the same time.

PR #792, opened 2019-05-20
feat(workspace): globbing support for root patterns

yatli:


chemzqm:
You should not use glob for this case, use `minimatch` instead when the pattern contains `*`.

yatli:
roger that.

src/util/fs.ts
```diff
@@ -5,6 +5,7 @@ import os from 'os'
 import path from 'path'
 import readline from 'readline'
 import util from 'util'
+import glob = require('glob')
```
chemzqm: No glob.

yatli: done.

yatli:
@chemzqm please take another look.

PR #810, opened 2019-05-25
Allow display of diagnostic info in Insert Mode

tombh:
Previously it was impossible to ever display diagnostic info when in
Insert Mode. There should at least be an option to do so.

Though, CursorHold behaviour will have to be set manually by the user
with:
`autocmd CursorHoldI * call CocActionAsync('diagnosticInfo')`

I couldn't see where to add this to the Chinese docs.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -428,6 +428,11 @@
       "description": "Join lines messages to reduce lines of floating window.",
       "default": false
     },
+    "diagnostic.showInInsertMode": {
+      "type": "boolean",
+      "description": "Allow viewing of diagnostic messages in Insert Mode, default false.",
+      "default": false
+    },
     "diagnostic.refreshOnInsertMode": {
       "type": "boolean",
       "description": "Enable diagnostic refresh on insert mode, default false.",
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -260,9 +260,19 @@ Builtin configurations:~
 	Join lines messages to reduce lines of floating window.,  default:
 	`false`
 
+"diagnostic.showInInsertMode":~
+
+	Allow display of diagnostics in insert mode, default false.,  default:
+	`false`
+
 "diagnostic.refreshOnInsertMode":~
 
 	Enable diagnostic refresh on insert mode, default false.,  default:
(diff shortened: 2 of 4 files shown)


chemzqm:
Just don't understand why this could be useful, since the language server will give you lost of errors when you're typing.

tombh:
This _only_ changes the ability to use the action: `diagnosticInfo` in Insert mode. All other behaviour is exactly the same as before.

Maybe you are confused with `diagnostic.refreshOnInsertMode`? Which already exists?

chemzqm:
It should always be allowed, no need for an option, just fixed.

tombh:
Great :)

tombh:
Thanks again for all the work you put into this plugin.

I noticed that diagnostic display in Insert Mode has regressed: https://github.com/neoclide/coc.nvim/blob/master/src/diagnostic/manager.ts#L456

Would it be ok to re-enable it? Thanks

chemzqm:
Fixed on latest release branch of coc.nvim.

tombh:
How did you fix it? I still cannot call the diagnostics window in Insert mode. I see that `echoMessage()` is still only allowed in Normal mode: https://github.com/neoclide/coc.nvim/blob/master/src/diagnostic/manager.ts#L456

Thank you!

tombh:
I still see `mode != n` on master:
```js
  public async echoMessage(truncate = false): Promise<void> {
    const config = this.config
    if (!this.enabled) return
    if (this.timer) clearTimeout(this.timer)
    let useFloat = config.messageTarget == 'float'
    let [bufnr, cursor, filetype, mode, disabled] = await this.nvim.eval('[bufnr("%"),coc#util#cursor(),&filetype,mode(),get(b:,"coc_diagnostic_disable",0)]') as [number, [number, number], string, string, number]
    if (mode != 'n' || bufnr == this.floatFactory.bufnr || disabled) return
    let diagnostics = this.getDiagnosticsAt(bufnr, cursor)
    if (diagnostics.length == 0) {
      if (useFloat) {
        this.floatFactory.close()
      } else {
...
```

PR #846, opened 2019-06-05
fix(diagnostic): correct ALE end column

rliebz:
An off-by-one error relating to the `end_col` field passed to ALE was causing an extra character to be highlighted.

Latest build, using ALE:
image

Latest build, using coc native highlighting (ignore color differences and note the underline range):
image

This pull request, using ALE:
image


chemzqm:
It should be problem of current highlight of coc.nvim.

PR #889, opened 2019-06-20
Fix comparing in getCurrentFunctionSymbol

liaoishere:
I've tried the newest release of coc for golang and typescript. The function `getCurrentFunctionSymbol` works well in typescript but is broken in golang. After a while for digging, I found the DocumentSymbols response from typescript server is different from golang.

Specifically, the `range` and `selectionRange` property of `DocumentSymbol` response has a different meaning. Here's the definitions from [LSP specification](https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol):

```
/**
* The range enclosing this symbol not including leading/trailing whitespace but everything else
 * like comments. This information is typically used to determine if the clients cursor is
 * inside the symbol to reveal in the symbol in the UI.
 */
range: Range;

/**
 * The range that should be selected and revealed when this symbol is being picked, e.g the name of a function.
 * Must be contained by the `range`.
 */
selectionRange: Range;
```

It says we should use `range` property to determine if the cursor is inside the symbol. While `selectionRange` is only the range of the symbol name string, not include the body.

So why the `selectionRange` also works for typescript? 

I thought that maybe another bug in `coc-tsserver`. I found the `range` and `selectionRange` property of DocumentSymbol response are exactly the same from `coc-tsserver`. But they are different in the response from gopls server. 
Anyway, We need to use `range` instead of `selectionRange` as the LSP specification told.

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -38,8 +38,8 @@ interface SymbolInfo {
   kind: string
   level?: number
   containerName?: string
-  selectionRange: Range
-  range?: Range
+  range: Range
+  selectionRange?: Range
 }
 
 interface CommandItem {
@@ -237,8 +237,8 @@ export default class Handler {
     let filetype = document.filetype
     let functionName = ''
     for (let sym of symbols.reverse()) {
-      if (sym.selectionRange
-        && positionInRange(position, sym.selectionRange) == 0
+      if (sym.range
+        && positionInRange(position, sym.range) == 0
         && !sym.text.endsWith(') callback')) {
         functionName = sym.text
         let kind = sym.kind.toLowerCase()
@@ -355,7 +355,7 @@ export default class Handler {
           text: name,
           level,
           kind: getSymbolKind(kind),
-          selectionRange: location.range,
+          range: location.range,
           containerName
         }
         res.push(o)


src/handler/index.ts
```diff
@@ -355,7 +355,7 @@ export default class Handler {
           text: name,
           level,
           kind: getSymbolKind(kind),
-          selectionRange: location.range,
+          range: location.range,
           containerName
```
iamcco: this is `SymbolInformation` so it should be `selectionRange` instead of `range` .

```
/**
	 * The location of this symbol. The location's range is used by a tool
	 * to reveal the location in the editor. If the symbol is selected in the
	 * tool the range's start information is used to position the cursor. So
	 * the range usually spans more then the actual symbol's name and does
	 * normally include things like visibility modifiers.
	 *
	 * The range doesn't have to denote a node range in the sense of a abstract
	 * syntax tree. It can therefore not be used to re-construct a hierarchy of
	 * the symbols.
	 */
	location: Location;
```
https://microsoft.github.io/language-server-protocol/specification#textDocument_documentSymbol

liaoishere: The spec says `So the range usually spans more than the actual symbol's name and does normally include things like visibility modifiers.` 

I've tested it on Python files with `coc-python` and jedi server. The range returned is obviously more than the symbol name. So I don't think this is equal to `selectionRange`

liaoishere: Here's a python LSP implementation, you can see the `SymbolInformation` result from the test file: https://github.com/palantir/python-language-server/blob/000993507a1bee471f4ab74f652e11875faed9a4/test/plugins/test_symbols.py#L44

The range of location covers the function name and body

iamcco: `selectionRange`:
- `The range that should be selected and revealed when this symbol is being picked`

`SymbolInformation`:
- `The location's range is used by a tool to reveal the location in the editor.`

They are more the same for usage. Also the spec didn't say `selectionRange` can't contain the function body.

liaoishere: You can also see the test file from javascript-typescript-langserver. In this `SymbolInformation` test: https://github.com/sourcegraph/javascript-typescript-langserver/blob/4c633f871faefb398f698bb555488d01ab33f93c/src/test/typescript-service-helpers.ts#L820. The range of symbol location starts from line 0, char 0 to line 0, char 33. It contains all the class definition code which is more than the symbol name. Test source is here: https://github.com/sourcegraph/javascript-typescript-langserver/blob/4c633f871faefb398f698bb555488d01ab33f93c/src/test/typescript-service-helpers.ts#L789

liaoishere: The spec says: 

`The range that should be selected and revealed when this symbol is being picked, e.g the name of a function.`

`selectionRange` is usually the range of the symbol name. You can go and check other exsiting LSP server implementations. A lot of examples:  Apple's swift LSP Server: https://github.com/apple/sourcekit-lsp/blob/20662cd553ae4b769055d55c6ee49ae52a8824cb/Tests/SourceKitTests/DocumentSymbolTests.swift#L101

liaoishere: Another example of `range` and `selectionRange` on a typescript LSP server: https://github.com/theia-ide/typescript-language-server/blob/d1a9a632a3e5e5ad5404a639ee7badcfb905d8f1/server/src/lsp-server.spec.ts#L246

chemzqm:
You're right, it should use range.

PR #913, opened 2019-06-25
Expose CodeAction functionality

kizza:
**Details**
- Specifically exposes `getCurrentCodeActions` and `applyCodeAction`
- The intention is to allow vim scripts, plugins or coc extensions to hook into this part of the plugin

```vim
let s:code_actions = coc#rpc#request('getCurrentCodeActions', [])

call coc#rpc#request('applyCodeAction', [s:code_actions[0]])
```

**Discussion**
- Happy to be pointed to where a better extension point might be to expose this functionality
- I didn't discover an obvious place to write tests which exercise which functions are exposed (ie. test the plugin interface). Happy to add something with some direction.
- I'm motivated towards these changes as it allows my popupmenu to [select and execute code actions](https://github.com/kizza/actionmenu.nvim/pull/1), and suspect others might be interested in being able to do similar things

(Thanks so much for `coc` it. is .awesome.)

### DIFF
--- a/src/plugin.ts
+++ b/src/plugin.ts
@@ -12,7 +12,7 @@ import listManager from './list/manager'
 import services from './services'
 import snippetManager from './snippets/manager'
 import sources from './sources'
-import { Autocmd, OutputChannel, PatternType } from './types'
+import { Autocmd, CodeAction, OutputChannel, PatternType } from './types'
 import clean from './util/clean'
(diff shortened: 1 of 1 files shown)


chemzqm:
It's already exits, checkout `:h coc-action-codeActions` and `:h coc-action-doCodeAction`

PR #946, opened 2019-07-05
Use package.json as source of truth for installed extensions

### LINKED ISSUE #945: g:coc_global_extensions doesn't seem to install if no existing extensions
## Result from CocInfo

vim version: NVIM v0.4.0-dev
node version: v10.5.0
coc.nvim version: 0.0.72-f7cd062403
term: iTerm.app
platform: darwin

## Describe the bug

Extensions are not installed if none exist. If I have no extensions, and I have `g:coc_global_extensions` defined, the extensions are not installed.
However the extensions are installed if I already have other extensions installed.

## Reproduce the bug

1. Remove `node_modules` from `coc/extensions`
2. Clear `package.json` dependencies from `coc/extensions/`
3. Make sure `g:coc_global_extensions` is defined
4. Open neovim

Current Behaviour:
* No extensions are installed
* If I manually install an extension, then all other missing extensions in `g:coc_global_extensions` are installed automatically

Expected behaviour:
* Extensions are installed automatically regardless of current extensions state ?

morgsmccauley:
On a fresh install of coc, the `extensions` directory is initialised with `package.json` and no `node_modules` folder. This causes the current code to silently fail when checking for currently installed extensions as it can't read the non existent `node_modules` folder.

I have changed the current behaviour to use `package.json` as the source of truth for installed extensions rather than `node_modules`. I think this is safer and faster as we don't need to loop the entire `node_modules` directory.

This also fixes #945 where global extensions will not be installed unless other extensions are already installed, i.e.`node_modules` exists.

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -161,16 +161,13 @@ export class Extensions {
   private async checkExtensions(): Promise<void> {
     let { globalExtensions, watchExtensions } = workspace.env
     if (globalExtensions && globalExtensions.length) {
-      let names = globalExtensions.filter(name => !this.isDisabled(name))
-      let folder = path.join(this.root, 'node_modules')
-      let files = await util.promisify(fs.readdir)(folder)
-      names = names.filter(s => files.indexOf(s) == -1)
+      let extensions = globalExtensions.filter(name => !this.isDisabled(name))
       let json = this.loadJson()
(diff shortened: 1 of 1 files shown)


morgsmccauley:
Wow, beat me to it, nice job :)

PR #974, opened 2019-07-11
Add config to set diagnostic filetype (syntax highlight diagnostics)

Avi-D-coder:
This allow you to apply either the parent buffers filetype or an arbitrary file type to diagnostic floats.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -550,16 +550,12 @@
       "description": "Maximum height of diagnostics floating window.",
       "default": 8
     },
-    "diagnostic.filetype": {
-      "type": "string",
-      "description": "The filetype given to diagnostic buffers. Use \"bufferType\" config to set diagnostics windows to parent code window's filetype.",
-      "default": "",
-      "enum": ["", "bufferType", "custom"]
-    },
     "diagnostic.filetypeMap": {
       "type":"object",
-      "description": "A map between buffer filetype and the filetype assigned to diagnostics. Use to override default \"diagnostic.filetype\"",
-      "default": {}
+      "description": "A map between buffer filetype and the filetype assigned to diagnostics. To syntax highlight diagnostics withs their parent buffer type use '\"default\": \"bufferType\"'",
+      "default": {
+        "default": ""
+      }
     },
     "signature.enable": {
       "type": "boolean",
--- a/data/schema.json
+++ b/data/schema.json
@@ -556,6 +556,11 @@
       "default": "",
       "enum": ["", "bufferType", "custom"]
     },
+    "diagnostic.filetypeMap": {
+      "type":"object",
+      "description": "A map between buffer filetype and the filetype assigned to diagnostics. Use to override default \"diagnostic.filetype\"",
+      "default": {}
+    },
     "signature.enable": {
       "type": "boolean",
       "description": "Enable signature help when trigger character typed, require restart service on change.",
--- a/data/schema.json
+++ b/data/schema.json
@@ -550,6 +550,12 @@
       "description": "Maximum height of diagnostics floating window.",
       "default": 8
     },
+    "diagnostic.filetype": {
+      "type": "string",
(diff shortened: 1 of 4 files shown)


chemzqm:
Why it's needed?

Avi-D-coder:
I find it hard to understand long intricate haskell/rust types without syntax highlighting.

oblitum:
> I find it hard to understand long intricate haskell/rust types without syntax highlighting.

But diagnostics output often is not valid code but more like prose with code excerpts.

chemzqm:
*  There could be multiple diagnostic messages in float window, so you should not use `setfiletype`
* `setfiletype` command can have some wired side effect like unexpected option change.

Avi-D-coder:
@chemzqm 
Multiple diagnostic messages don't break the any more than one. The color differentiation between diagnostic levels is removed if you use this option, so nothing else breaks.

what options can change? What should I guard against?

Avi-D-coder:
@oblitum 

Even with simple errors and some breakage I find it's worth it. It's off by default so no one will experience incorrect highlights unless they turn it on.

Once treesitter lands these sorts of hack won't be needed, but until then they are useful.

Screenshot_20190711_171747

chemzqm:
Looks like the configuration should be a map that used for map the filetypes of different buffers.

`set filetype` is not needed on neovim since the highlight already added by `floatBuffer` module which use background neovim for highlights.

Avi-D-coder:
@chemzqm I was originally going to make config be mappings, but It added a lot of complexity for the user, when almost all cases are handled by the special option "bufferType" or "markdown". To be ergonomic a fallback/default filetype config would still need to be preset and that's what we already have.

I am hoping this feature becomes subsumed by tree-sitter, which would need no such mapping.

chemzqm:
>  but It added a lot of complexity for the user

It's only useful for specific filetypes, so I think it should only work for the filetypes specified in the map, instead make all diagnostic buffer have filetype which can be easily broken.

> I am hoping this feature becomes subsumed by tree-sitter

It won't help for this case, the filetype of diagnostic bufffer should limited by filetypes provided by user.

Avi-D-coder:
@chemzqm Okay final got around to this. Should be ready

Avi-D-coder:
@chemzqm could you take a look at this again?

chemzqm:
I don't think `diagnostic.filetype` is needed when there is `diagnostic.filetypeMap`.
`diagnostic.filetype` would cause all diagnotic buffer have different filetype and lose highlight of error and wanring which should be default.

Avi-D-coder:
I don't want to have to update `diagnostic.filetypeMap` every time I use new language, for people like me changing the default is useful.
Otherwise I'll be left with an ever expanding version of this:
```json
	"diagnostic.filetypeMap": {
		"haskell": "haskell",
		"rust": "rust",
		"cpp": "cpp",
		"swift": "swift",
		"go": "go",
		"scala": "scala"
	},
```

chemzqm:
You can add support for regex pattern like:
```
"diagnostic.filetypeMap": {
		"(.*)": "$1",
	},
```
two different configuration for same is confusing.

Avi-D-coder:
Regex is overkill. I suspect most users will want **all** diagnostics set to either "", or  "bufferType". The override map use case is nice to have, the more import one is the simple default config.
What about `"diagnostic.defaultFiletype": "bufferType"`?

Or this, but if someone defines a vim filetype default, this will break.
```
"diagnostic.filetypeMap": {
		"default": "bufferType",
},
```

Avi-D-coder:
@chemzqm I would love to get a form of default merged.


I very much doubt anyone will make a file type called `AllFileTypesDefault`, so this should't have any major  drawbacks. Would you merge this?
```json
"diagnostic.filetypeMap": {
		"AllFileTypesDefault": "bufferType",
},
```

chemzqm:
The default key looks good to me.

Avi-D-coder:
@chemzqm How's this?

PR #989, opened 2019-07-17
Added support for cygwin

cosminadrianpopescu:
Right now, in cygwin, the vim process will return via the `getpid()` command the cygwin pid. When trying to check that process from the node process (which is not a posix process) that will fail and after 15 minutes the coc.nvim is closed. 

Via this pull request, I am trying to get the winpid, instead of the cygwin pid. 

However, I was wondering what is the purpose of having the `workspace.ts::checkProcess` logic? The process launched via `job_start` in vim is a child process of vim. So, anyway, as soon as the vim process does not exists anymore, the node process will be killed automatically by the OS. This is valid for Linux, Mac and cygwin. I don't know for windows. But if this is not valid for windows, can we modify the function to be called only for Windows?

There is really no use for the timer here:

https://github.com/neoclide/coc.nvim/blob/3ce5cabf4c83f40fa7557d65d0e718519cc36422/src/workspace.ts#L1597

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -551,6 +551,15 @@ function! coc#util#run_terminal(opts, cb)
   call coc#util#open_terminal(opts)
 endfunction
 
+function! coc#util#getpid()
+  if !has('win32unix')
+    return getpid()
+  endif
+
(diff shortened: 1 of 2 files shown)


chemzqm:
Thanks a lot.

PR #997, opened 2019-07-18
Fixed EOL for cygwin

cosminadrianpopescu:
When running under cygwin, if your files are using unix line ending, because of the node process being a Window process, there will be windows line endings added when modifying code via coc.nvim (for example when adding a new import via `:CocFix`). 

This PR aims to fix this. It checks if the vim is running under cygwin and if the node process is running under windows and if the fileformat is unix, then it will replace the end lines with linux end lines.


chemzqm:
It's bug of your language server, it should not include `\r` in TextEdit.
Fixed by replace all `\r` in TextEdit.

PR #1005, opened 2019-07-20
Fixed the paths received from a windows node process to a cygwin vim …

cosminadrianpopescu:
…process

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -212,7 +212,11 @@ function! coc#util#execute(cmd)
 endfunction
 
 function! coc#util#jump(cmd, filepath, ...) abort
-  let file = fnamemodify(a:filepath, ":~:.")
+  let path = a:filepath
+  if (has('win32unix'))
+    let path = substitute(a:filepath, '\v\\', '/', 'g')
+  endif
+  let file = fnamemodify(path, ":~:.")
   if a:cmd =~# '^tab'
     exe a:cmd.' '.fnameescape(file)
     if !empty(get(a:, 1, []))
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -288,6 +288,7 @@ function! coc#util#get_bufoptions(bufnr) abort
   let bufname = bufname(a:bufnr)
   return {
         \ 'bufname': bufname,
+        \ 'isCygwin': has('win32unix'),
         \ 'eol': getbufvar(a:bufnr, '&eol'),
         \ 'variables': s:variables(a:bufnr),
         \ 'fullpath': empty(bufname) ? '' : fnamemodify(bufname, ':p'),
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -288,7 +288,6 @@ function! coc#util#get_bufoptions(bufnr) abort
   let bufname = bufname(a:bufnr)
   return {
         \ 'bufname': bufname,
-        \ 'isCygwin': has('win32unix'),
         \ 'eol': getbufvar(a:bufnr, '&eol'),
         \ 'variables': s:variables(a:bufnr),
         \ 'fullpath': empty(bufname) ? '' : fnamemodify(bufname, ':p'),
@@ -581,6 +580,7 @@ function! coc#util#vim_info()
         \ 'completeOpt': &completeopt,
         \ 'pumevent': exists('##MenuPopupChanged') || exists('##CompleteChanged'),
         \ 'isVim': has('nvim') ? v:false : v:true,
+        \ 'isCygwin': has('win32unix') ? v:true : v:false,
         \ 'isMacvim': has('gui_macvim') ? v:true : v:false,
         \ 'colorscheme': get(g:, 'colors_name', ''),
         \ 'workspaceFolders': get(g:, 'WorkspaceFolders', v:null),
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -115,7 +115,7 @@ export default class Document {
     this._changedtick = opts.changedtick
     this._rootPatterns = opts.rootPatterns
     this.eol = opts.eol == 1
-    let uri = this._uri = getUri(opts.fullpath, buffer.id, buftype, opts.isCygwin)
+    let uri = this._uri = getUri(opts.fullpath, buffer.id, buftype, this.env.isCygwin)
     token.onCancellationRequested(() => {
       this.detach()
     })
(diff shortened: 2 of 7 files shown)


autoload/coc/util.vim
```diff
@@ -288,6 +288,7 @@ function! coc#util#get_bufoptions(bufnr) abort
   let bufname = bufname(a:bufnr)
   return {
         \ 'bufname': bufname,
+        \ 'isCygwin': has('win32unix'),
```
chemzqm: It should add to vim info at https://github.com/neoclide/coc.nvim/blob/master/autoload/coc/util.vim#L562, then use `workspace.env.isCygwin`

src/model/document.ts
```diff
@@ -115,7 +115,7 @@ export default class Document {
     this._changedtick = opts.changedtick
     this._rootPatterns = opts.rootPatterns
     this.eol = opts.eol == 1
-    let uri = this._uri = getUri(opts.fullpath, buffer.id, buftype)
+    let uri = this._uri = getUri(opts.fullpath, buffer.id, buftype, opts.isCygwin)
```
chemzqm: Use `isCygwin` from env instead.

src/types.ts
```diff
@@ -390,6 +390,7 @@ export interface BufferOption {
   changedtick: number
   rootPatterns: string[] | null
   additionalKeywords: string[]
+  isCygwin: boolean
```
chemzqm: Move to `Env` interface.

cosminadrianpopescu: In Env is not a good option, since getUri will be called also before the workspace is ready. 

So I thought is maybe better to have it through buffer variables. But now I think it would be best to have it as nvim.eval('has("win32unix")'). 

What do you think?

cosminadrianpopescu:
I am also thinking about adding a `normalize_path` function in the util, which should be passed also to the daemon. This function should be activated via an option in the config (something like `g:coc_normalize_path_function`. 

There are still issues with the paths in some places, like for example the fact that `/cygdrive/c` sometimes does not exists. And then, if I create the `c:/cygdrive` folder and inside it I make a junction towards `c:/`, now I have a `c:/cygdrive/c` which exists, but then I would have files returned by the node daemon like `c:/cygdrive/c/cygdrive/c/Users/myUser/coc.nvim/src/workspace.ts`, which is again a problem. 

So, to fix this, I am thinking to this option. What do you think?

src/types.ts
chemzqm: > since getUri will be called also before the workspace is ready.

All documents are created after `env` initialized, which is passed to document on create, so it's not a problem.

chemzqm:
I don't like `g:coc_normalize_path_function`, it can easy goes wrong and make things more complicated.
For that issue, the javascript code could be patched to not create path like `c:/cygdrive/c/cygdrive/c/...`

cosminadrianpopescu:
> For that issue, the javascript code could be patched to not create path like c:/cygdrive/c/cygdrive/c/...

Do you have any ideas at which level would it be easier? 

At the level of webpack_require? If so, what import I should override? `vscode-uri`? 

At the level of server.js? In which case, then I would be forced to only using the compiled version (which would not be such a big deal for me, but if I can, I prefer to use the pack bundle).

Or are you actually thinking of something else?

chemzqm:
Find out the module that create invalid path.

cosminadrianpopescu:
> Find out the module that create invalid path.

That's again easier said than done. Because probably it's in the node `fs` module. Meanwhile, I patched recently the `server.js` file and it worked perfectly since then.

I can understand the reluctance to have a `normalize_path_function`. But the plugin will not work properly in cygwin. 

I am proposing something similar with what I've done in the server.js. In the `./util/fs.ts`:

```
const URI = require('vscode-uri')
const old_uri = URI.URI.file;
const isCygwin = nvim.eval('has("win32unix")');
const prefixes = nvim.eval('get(g:, 'coc_cygwin_path_prefixes');
URI.URI.file = function(_path) {
  if (!isCygwin || !platform.isWindows) {
    return old_uri(_path);
  }
  let path = _path.replace(/\\/g, '/');
  Object.keys(prefixes).forEach(k => path = path.replace(new RegExp(`^${k}`, 'g'), prefixes[k]);
  return old_uri(path);
}

```

And the `g:coc_cygwin_path_prefixes` variable would be a dictionary where the key would be the node path and the value would be the vim path. In my case, it would look something like this:

```
{
  '/cygdrive/c: '',
  '/cygdrive/c/home': '/home',
  '/cygdrive/c/cygdrive/c': '',
  '/cygdrive/c/Users/myUser': '/home',
}
```

What do you think about this? This would make it easier to identify and locate possible bugs, since this variable would by by default un-set.

Unfortunately I could not make it works 100% without something like this. Bellow you will find my patch of the `server.js` file (this is specific to my case): 

```
const URI = require('vscode-uri')
var old_uri = URI.URI.file;
URI.URI.file = function(_path) {
  let path = _path.replace(/\\/g, '/');
  path = path.replace(/^c:/g, '/cygdrive/c')
  path = path.replace(/^\/cygdrive\/c\/Users\/myUser/g, '/home/myUser')
  const p = /^\/cygdrive\/c/g
  while (path.match(p)) {
    path = path.replace(p, '');
  }
  path = path.replace(/^\/home\/myUser/g, '');
  path = '/home/myUser' + path;
  return old_uri(path);
}
```

This basically ensures that all my paths will begin with `/home/myUser`, which solves all my problems. As I was saying, without this I could not have it working 100%. 

I am waiting for your answer.

chemzqm:
> I am proposing something similar with what I've done in the server.js. In the ./util/fs.ts:

You won't have nvim instance in `util/fs.ts`, you can add that code to `src/attach.ts` just after attach.

chemzqm:
Found some lint issues:
```
$ tslint -c tslint.json -p .
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:15:3 - Promises must be handled appropriately
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:17:13 - Unnecessary semicolon
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:19:29 - Unnecessary semicolon
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:20:5 - Promises must be handled appropriately
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:21:18 - non-arrow functions are forbidden
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:22:45 - Unnecessary semicolon
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:23:104 - Unnecessary semicolon
ERROR: /home/travis/build/neoclide/coc.nvim/src/attach.ts:24:29 - Unnecessary semicolon
```

PR #1014, opened 2019-07-23
fix(language-client): mismatch createMessageTransports type

bmatcuk:
I've updated the type of BaseLanguageClient.createMessageTransports to match the type of LanguageClient.createMessageTransports. In my project, using typescript, I was getting compile errors because the types didn't match.

### DIFF
--- a/src/language-client/client.ts
+++ b/src/language-client/client.ts
@@ -3782,7 +3782,7 @@ export abstract class BaseLanguageClient {
 
   protected abstract createMessageTransports(
     encoding: string
-  ): Thenable<MessageTransports>
+  ): Promise<MessageTransports | null>
 
   private createConnection(): Thenable<IConnection> {
     let errorHandler = (error: Error, message: Message, count: number) => {


chemzqm:
Use `await Promise.resolve` instead

bmatcuk:
I'm confused what you mean... This PR changes the return value of createMessageTransports so that BaseLanguageClient matches LanguageClient. Here's the error I get *without* this change:

```
node_modules/coc.nvim/lib/language-client/index.d.ts:84:15 - error TS2416: Property 'createMessageTransports' in type 'LanguageClient' is not assignable to the same property in base type 'BaseLanguageClient'.
  Type '(encoding: string) => Promise<MessageTransports | null>' is not assignable to type '(encoding: string) => Thenable<MessageTransports>'.
    Type 'Promise<MessageTransports | null>' is not assignable to type 'Thenable<MessageTransports>'.
      Types of property 'then' are incompatible.
        Type '<TResult1 = MessageTransports | null, TResult2 = never>(onfulfilled?: ((value: MessageTransports | null) => TResult1 | PromiseLike<TResult1>) | null | undefined, onrejected?: ((reason: any) => TResult2 | PromiseLike<...>) | null | undefined) => Promise<...>' is not assignable to type '{ <TResult>(onfulfilled?: ((value: MessageTransports) => TResult | Thenable<TResult>) | undefined, onrejected?: ((reason: any) => TResult | Thenable<TResult>) | undefined): Thenable<TResult>; <TResult>(onfulfilled?: ((value: MessageTransports) => TResult | Thenable<...>) | undefined, onrejected?: ((reason: any) => v...'.
          Types of parameters 'onfulfilled' and 'onfulfilled' are incompatible.
            Type '((value: MessageTransports) => any) | undefined' is not assignable to type '((value: MessageTransports | null) => any) | null | undefined'.
              Type '(value: MessageTransports) => any' is not assignable to type '(value: MessageTransports | null) => any'.
                Types of parameters 'value' and 'value' are incompatible.
                  Type 'MessageTransports | null' is not assignable to type 'MessageTransports'.
                    Type 'null' is not assignable to type 'MessageTransports'.

84     protected createMessageTransports(encoding: string): Promise<MessageTransports | null>;
                 ~~~~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```

bmatcuk:
To be clear, I'm not even using createMessageTransports in my code! In fact, at this point, my code is just a super simple LanguageClient. But it won't compile because LanguageClient extends BaseLanguageClient, and the type of createMessageTransports does not match between the two.

chemzqm:
`Thenable` is compatible with `Promise` and I don't understand why you need compile coc.nvim with your project since it's already compiled when using from npm.

bmatcuk:
From the error message: `Type 'MessageTransports | null' is not assignable to type 'MessageTransports'.` Nothing to do with Thenable vs Promise, it's the `| null` bit that's the problem. If it'll make you happy, I can switch it back to Thenable in my PR, though, the signatures between the two functions would not be exactly the same.

I'm not compiling coc.nvim as part of my project. But because I'm importing coc.nvim, typescript needs to compile bits of coc.nvim into my project, and that's where it's complaining.

Here's a minimal example to show the issue: https://gist.github.com/bmatcuk/9419b85abd22c123c3baa865b49ee04b

Place all three files in the same directory and run `yarn install` and then `yarn run build`.

The problem appears to be the typescript option "strictNullChecks". If I remove it from my config, it compiles. But I don't want to remove it from my config because 1) I like strict null checks, and, 2) if I remove it, typescript will also force me to remove the "strictPropertyInitialization" option, which I'm also a fan of, because it depends on strict null checks.

chemzqm:
Ok, I have added `null`

bmatcuk:
ok, that works.

Issue #1022, opened 2019-07-24
Create files with CocList


shtanton:
**Is your feature request related to a problem? Please describe.**
As far as I can tell, the actions on folders are rename and delete, it would be nice to be able to create a new file in a folder from CocList.

**Describe the solution you'd like**
Select a folder and then action:addfile or something to create the file

**Describe alternatives you've considered**
Currently just :e path/to/new/file.ext which is probably the only file interaction I haven't replaced with lists.

Apologies if this already exists and I just couldn't find it. If it doesn't, I'd be happy to write a PR for it.

chemzqm:
Ok, consider send a PR for it.

PR #1028, opened 2019-07-26
[RFC] Allows creating files from folders list

shtanton:
For issue #1022 
Adds 'newfile' action on folders list

### DIFF
--- a/src/list/source/folders.ts
+++ b/src/list/source/folders.ts
@@ -3,11 +3,9 @@ import { statAsync } from '../../util/fs'
 import { ListContext, ListItem } from '../../types'
 import workspace from '../../workspace'
 import BasicList from '../basic'
-import {Range, Location} from 'vscode-languageserver-protocol'
 import {URI} from 'vscode-uri'
 import mkdirp from 'mkdirp'
 import path from 'path'
-import fs from 'fs'
 
 export default class FoldList extends BasicList {
   public defaultAction = 'edit'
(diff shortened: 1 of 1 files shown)


shtanton:
I'm not sure if you want tests for the new action, I don't know how to add them as the only existing test for folders seems to be checking the delete action runs, but not checking it deletes anything.

If you want some unit tests I'll see if I can figure it out

src/list/source/folders.ts
```diff
@@ -25,6 +30,18 @@ export default class FoldList extends BasicList {
     this.addAction('delete', async item => {
       workspace.removeWorkspaceFolder(item.label)
     }, { reload: true, persist: true })
+
+		this.addAction('newfile', async item => {
+			let file = await nvim.call('input', ['File name:', item.label + '/'])
```
chemzqm: You have to use `workspace.requestInput` api instead,  this call won't work on vim8.

shtanton: Cool, I've fixed that

src/list/source/folders.ts
```diff
@@ -25,6 +30,18 @@ export default class FoldList extends BasicList {
...
+		this.addAction('newfile', async item => {
+			let file = await workspace.requestInput('File name:', item.label + '/')
+			let dir = path.dirname(file)
+			if (!fs.existsSync(dir)) {
+				mkdirp.sync(dir)
+			}
+			await workspace.createFile(file, {overwrite: false, ignoreIfExists: true})
+			let range = Range.create(0, 0, 0, 0)
+			let location = Location.create(URI.file(file).toString(), range)
+			await this.jumpTo(location)
```
chemzqm: No need for location, Just use `await this.jumpTo(URI.file(file).toString())`

shtanton: Thanks, updated that

PR #1077, opened 2019-08-11
No delay for echoed diagnostics, configurable delay for floating diagnostics

tbodt:
image

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -467,6 +467,11 @@
       "default": "float",
       "enum": ["echo", "float"]
     },
+    "diagnostic.floatDelay": {
+      "type": "number",
+      "description": "How long to wait (in milliseconds) before displaying the diagnostics floating window.",
+      "default": 500
+    },
     "diagnostic.joinMessageLines": {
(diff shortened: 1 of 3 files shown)


tbodt:
Realizing that this is actually useful if your diagnostics are floating. I'll make it keep the timeout in that case, and also make it a config option.

chemzqm:
The delay is inteded to avoid unnecessary message when jump quickly like use vim command

tbodt:
What do you mean by jumping quickly? I want the echoed diagnostic to always update immediately when I move my cursor.

tbodt:
Oh are you talking about the hit-enter prompts? I think I can get rid of those.

chemzqm:
It also cause bad performance when you move cursor in error region, so no hope.

chemzqm:
You can use jump keymappings to have messages shown immediately.

tbodt:
Can I make the delay configurable at least?

chemzqm:
Sure, you can create a PR for it.

Issue #718, opened 2019-04-22
Location of the coc-settings.json file


nostdm:
**Problem**

At the moment `coc` configuration files can be found in `$HOME/.config/coc` and `$HOME/.config/nvim`. As far as I know there is no option to change the location of the settings file. Since I like to keep my `$HOME/.config/nvim` directory clean and organised, I would like to be able to put all the `coc` related configuration inside `$HOME/.config/coc` and keep `$HOME/.config/nvim` for vimscript configuration only. 

**Feature request**

Provide a way to override the default configuration file location either in the `init.vim` by setting a global variable e.g. `let g:coc_config_file="$HOME/.config/coc/settings.json"` or as a command line flag when running the `coc` server e.g. `--config="$HOME/.config/coc/settings.json"`.

P.S. The name `coc-settings.json` is a bit awkward - IMO it should be called either `coc/settings.json` or `coc.json` ideally, the user could specify the config file name without losing config autocompletion.

chemzqm:
No, you can create a link or use `g:coc_user_config` or `coc#config()` in viml.

nostdm:
When using `coc#config()` I would have to define the JSON configuration inside a vimscript file, which doesn't have linting and syntax support as far as I know. I understand that `coc-settings.json` is used for workspace configuration but having a variable for global user config would not impact this `let g:coc_global_config="$HOME/.config/coc/coc-settings.json"`. Or at least do you mind explaining why you decided to put the file inside the `nvim` directory since that directory is generally used for vimscript configuration. Thanks.

yatli:
@nostdm no it does not imply you have to place the content in vimL.
vimL is capable of file reading, so if you need the capability of dynamic/conditional config you can use vimL to branch `coc-settings.json` in different subfolders, which won't be picked up by default, and you'll still have all the nice linting and auto completion.

coldspark29:
So where is this frigging file now? Can't find it

chemzqm:
https://github.com/neoclide/coc.nvim/wiki/Using-the-configuration-file

antonwnk:
@coldspark29 
Behold the glorious `:CocConfig` command.

kaungmyathein12:
@antonwnk Thank you.

PR #1103, opened 2019-08-21
Allow coc_config_home to be configurable

aste4:
This is something that has been asked about in the past - one example (https://github.com/neoclide/coc.nvim/issues/718)

This helps improve maintenance of dotfiles for both vim and neovim. It allows for a way to use coc.nvim with neovim without needing to add a symbolic link somewhere to to .config. Everything is configurable from within the vimrc.

Let me know your thoughts.

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -347,6 +347,9 @@ function! coc#util#preview_info(info, ...) abort
 endfunction
 
 function! coc#util#get_config_home()
+  if exists("g:coc_config_home")
+      return get(g:, 'coc_config_home', "")
+  endif
   if exists('$VIMCONFIG')
     return resolve($VIMCONFIG)
   endif
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -347,8 +347,8 @@ function! coc#util#preview_info(info, ...) abort
 endfunction
 
 function! coc#util#get_config_home()
-  if exists("g:coc_config_home")
-      return get(g:, 'coc_config_home', "")
+  if !empty(get(g:, 'coc_config_home', ''))
+      return g:coc_config_home
   endif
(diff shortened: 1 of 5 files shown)


autoload/coc/util.vim
```diff
@@ -347,6 +347,9 @@ function! coc#util#preview_info(info, ...) abort
 endfunction
 
 function! coc#util#get_config_home()
+  if exists("g:coc_config_home")
```
chemzqm: You should use `if !empty(get(g:, 'coc_config_home', ''))`

aste4: It should be fixed. Let me know if there is anything else

chemzqm:
Name of environment variable VIMCONFIG used by node process also need changed to avoid confilct issue.

aste4:
Why does it need to be renamed? What is the conflict issue? 

I am looking at where it is used and can't think of anything going wrong. It does not seem to be a reserved or well known environment variable name. 

If it really needs to be renamed, what name would you suggest? 

from `VIMCONFIG` to `COC_VIMCONFIG`? Wouldn't the conflict issue still exist?

chemzqm:
In rpc.vim

aste4:
I see it in rpc.vim. But I don't understand what it has to do with this PR. 

Do you mean to rename `VIMCONFIG` to `COC_VIMCONFIG` in rpc.vim, workspace.ts and extensions.ts? With that change `coc_config_home` is then unnecessary as I can just set VIMCONFIG in the vimrc before loading coc.nvim. But it is easier to have a user settable global variable instead

chemzqm:
You can use VIMCONFIG without your PR.
The problem is if user use that global variable VIMCONFIG would be changed.

aste4:
Thank you.  You are correct. 

It should now be fixed.

PR #1110, opened 2019-08-23
add :CocLocalConfig command

amiralies:


chemzqm:
Name of environment variable VIMCONFIG used by node process also need changed to avoid confilct issue.

voldikss:
@amiralies 

If I am under the **sub-directory** of the **root** path, it should  search upwards for the `.vim` folder.

amiralies:
good idea i've just imeplemented this the way vscode behaves with localconfig,
i'm with you, i think it's better to recursivly search parents for `.vim` until `home`

plugin/coc.vim
```diff
@@ -94,6 +94,22 @@ function! s:OpenConfig()
   execute 'edit '.home.'/coc-settings.json'
 endfunction
 
+function! s:OpenLocalConfig()
+  let currentDir = $PWD
```
chemzqm: Should use `getcwd()` instead of `$PWD`

plugin/coc.vim
```diff
@@ -94,6 +94,22 @@ function! s:OpenConfig()
...
+
+  while isdirectory(currentDir) && !(currentDir ==# $HOME) && !(currentDir ==# fsRootDir)
+    if isdirectory(currentDir.'/.vim')
+      execute 'edit '.currentDir.'/.vim/coc-settings.json'
+      return
+    endif
+    let currentDir = fnamemodify(currentDir, ':p:h:h')
+  endwhile
+
+  call mkdir('.vim', 'p')
```
chemzqm: Should we prompt user for create `.vim` folder? Some user would be confused with it.

doc/coc.txt
```diff
@@ -1567,6 +1567,10 @@ COMMANDS 					*coc-commands*
 
 		Edit the user config file `coc-settings.json`
 
+:CocLocalConfig 				*:CocLocalConfig*
+
+		Edit the workspace config file `.vim/coc-settings.json`
```
chemzqm: It's not very clear, should be: edit or create `.vim/coc-settings.json` of current working directory.

amiralies: You're right. current working directory or current workspace since it search upwards for `.vim/coc-settings.json`?

chemzqm: There're workspaceFolders but no workspace like VSCode with coc.nvim.
For a single file, the folder contains `.vim/settings.json` could be different from it's workspaceFolder.
Could be better if `vim/settings.json` can be resolved from workspaceFolder of current buffer or parent directory of current buffer, I'm ok with cwd since it's same for my use case.

plugin/coc.vim
amiralies: You mean sth like this?
```vim
  if coc#util#prompt_confirm("No local config detected, would you like to create .vim/coc-settings.json?")
    call mkdir('.vim', 'p')
    execute 'edit .vim/coc-settings.json'
  endif
```

chemzqm: Yes.

amiralies: What if cwd is `$HOME` should we return `OpenConfig`?

chemzqm: I suppose show warning message: Can't resolve local config from current working directory.

PR #1150, opened 2019-09-06
feat(languages): formatOnSave for all filetypes

### LINKED ISSUE #1149: [Question] Automatically Format all file types?
Is there any way to tell Coc.nvim to format *all* file types on save? `coc.preferences.formatOnSaveFiletypes` takes a list of file types as argument, and it would be nice if I could do something like this `"coc.preferences.formatOnSaveFiletypes": ["*"]` to enable it for all file types, instead of having to type them all out.

fannheyward:
close #1149

### DIFF
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -109,7 +109,7 @@ class Languages {
       let { languageId } = event.document
       let config = workspace.getConfiguration('coc.preferences', event.document.uri)
       let filetypes = config.get<string[]>('formatOnSaveFiletypes', [])
-      if (filetypes.indexOf(languageId) !== -1 || filetypes === ['*']) {
+      if (filetypes.indexOf(languageId) !== -1 || filetypes.some(item => item === '*')) {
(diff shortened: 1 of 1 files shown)


src/languages.ts
```diff
@@ -109,7 +109,7 @@ class Languages {
       let { languageId } = event.document
       let config = workspace.getConfiguration('coc.preferences', event.document.uri)
       let filetypes = config.get<string[]>('formatOnSaveFiletypes', [])
-      if (filetypes.indexOf(languageId) !== -1) {
+      if (filetypes.indexOf(languageId) !== -1 || filetypes === ['*']) {
         let willSaveWaitUntil = async (): Promise<TextEdit[]> => {
```
iamcco: `filetypes === ['*']` will never be `true` in this way `===`.

fannheyward: Fixed, toooooo stupid mistake.

PR #1169, opened 2019-09-12
update example configuration

aca:
Use complete_info instead of pumvisible to determine whether new
line should start or not. Current mapping requires two cr key press
to get to newline when completion item is not selected.

I found it really annoying, but if it's intended behavior just close it.
Thanks for the great project. I should say that it literally changed my life.


chemzqm:
`complete_info` function could be not exists on old vim/neovim, I've improved it.

PR #1205, opened 2019-09-24
fix(api.vim): vim no support v:true, v:false as option value

weirongxu:


chemzqm:
Looks like https://github.com/neoclide/coc.nvim/blob/master/autoload/coc/api.vim#L465 need to be changed as well.

weirongxu:
OK, I have changed it

PR #1218, opened 2019-09-26
Update Readme.md

jrowlingson:
Floating windows are now included in neovim as of the [`0.4.0`](https://github.com/neovim/neovim/releases/tag/v0.4.0) release 🎉

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -25,7 +25,7 @@ It's a completion framework and language server client which supports [extension
 
 _True snippet and additional text editing support_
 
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -17,15 +17,15 @@
 
 Coc is an intellisense engine for vim8 & neovim.
 
-It works on `vim >= 8.0` and `neovim >= 0.3.1`.
+It works on `vim >= 8.0` and `neovim >= 0.4.0`.
```
fannheyward: Coc can work on Neovim < 0.4, just without floating window support. So, there is no requirement must be 0.4+.

jrowlingson: Thanks - I've reverted this change and added a note that floating windows are supported in neovim `0.4.0`.

oblitum:
> nightly build of vim >= 8.1.1522

This term is wrong, as larger Vim versions have been available in some package managers officially for a long time. It's not nightly build.

PR #1224, opened 2019-09-28
feat(diagnostic): add maxWindowWidth config

todesking:


svanharmelen:
Any idea when we can expect this in the release branch? Or is it OK to use master if I want to use this feature? Thanks!

PR #1236, opened 2019-10-02
fix: diagnostic.refreshOnInsertMode

clyfish:
1. `diagnostic.refreshOnInsertMode` should work when `diagnostic.displayByAle == true`
2. `diagnostic.refreshOnInsertMode` should work when `diagnostics.length == 0`

### DIFF
--- a/src/diagnostic/manager.ts
+++ b/src/diagnostic/manager.ts
@@ -551,12 +551,12 @@ export class DiagnosticManager implements Disposable {
 
   private refreshBuffer(uri: string): boolean {
     let { insertMode } = workspace
+    if (insertMode && !this.config.refreshOnInsertMode) return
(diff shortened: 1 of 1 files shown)


chemzqm:
> diagnostic.refreshAfterSave should work when diagnostic.displayByAle == true

No, you should use ale to configure it.

clyfish:
I can't use ale to configure it.

My ale config:
```
let g:ale_lint_on_text_changed = 'never'
let g:ale_open_list = 1
```

clyfish:
> It's intended to clear diagnostics earlier.

If `diagnostic.refreshOnInsertMode` is true, the diagnostics should be cleared when leaving insert mode.

chemzqm:
You're right.

PR #1262, opened 2019-10-11
feat(completion): add suggest.completionDefaultSortMethod

### LINKED ISSUE #1070: user configuration for sorting completion items?
currently the completion is sorted by text lengths, which is what i dislike. i want them sort in alphabetical ordering. i have to modify `completion.ts` in this way:
`return a.filterText.length - b.filterText.length` =>
`return a.filterText.localeCompare(b.filterText)`

i think there should be an option to specify which order the completion items are sorted in. for example: let `g:coc_completion_order="alphabetic"` or something similar in coc-settings.json.

**Additional context**

amiralies:
added option to change default behavior for sorting
which is by length.
there is currently one other option `alphabetical`

Closes #1070

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -416,6 +416,12 @@
       "description": "Remove completion items with duplicated word for all sources, snippet items are excluded.",
       "default": false
     },
+    "suggest.completionDefaultSortMethod": {
+      "type": "string",
+      "description": "Default sorting behavior for suggested completion items.",
+      "default": "length",
+      "enum": ["length", "alphabetical"]
+    },
     "suggest.completionItemKindLabels": {
       "type": "object",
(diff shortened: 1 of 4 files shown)


data/schema.json
```diff
@@ -416,6 +416,12 @@
       "description": "Remove completion items with duplicated word for all sources, snippet items are excluded.",
       "default": false
     },
+    "suggest.completionDefaultSortMethod": {
```
fannheyward: Thanks for your PR. Maybe it's better to named `suggest.defaultSortMethod`, as suggest already means completion.

amiralies:
Good catch!

fannheyward:
cc @chemzqm

PR #1286, opened 2019-10-20
Fixes the usage of documentChanges

sawmurai:
This PR fixes the usage of `documentChanges`. Previously `if (!changes)` could never be true because right in the line before `changes` is initialized with `{}` as a fallback.

This - at least for me - broke the php language server.


sawmurai:
I think the failing tests are unrelated to my change

marlonfan:
cool!

PR #1287, opened 2019-10-21
feat(diagnostic): classify ALE items by their source

yous:
With coc-python extension, pylint and flake8 enabled and `displayByAle` is on, say we have this file:
``` python
def foo():
    print('foo')

def bar():
    print('bar')


var = baz
```

Then there are error messages:
```
test.py|4 col 1 error| [python] E302: expected 2 blank lines, found 1
test.py|8 col 7 error| [python] F821: undefined name 'baz'
test.py|8 col 7 error| [python] E0602: Undefined variable 'baz' (undefined-variable)
```

Note that they all have `[python]` header. This patch changes this to:
```
test.py|4 col 1 error| [flake8] E302: expected 2 blank lines, found 1
test.py|8 col 7 error| [flake8] F821: undefined name 'baz'
test.py|8 col 7 error| [pylint] E0602: Undefined variable 'baz' (undefined-variable)
```
And this behavior is more similar to ALE's one.

### DIFF
--- a/src/diagnostic/manager.ts
+++ b/src/diagnostic/manager.ts
@@ -570,19 +570,28 @@ export class DiagnosticManager implements Disposable {
         if (level) {
           diagnostics = diagnostics.filter(o => o.severity && o.severity <= level)
         }
-        let aleItems = diagnostics.map(o => {
-          let { range } = o
-          return {
-            text: o.message,
-            code: o.code,
+        let sourceItemsMap = diagnostics.reduce((prev, cur) => {
+          let { range } = cur
+          let items = prev.get(cur.source)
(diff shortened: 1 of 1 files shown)


yous:
Ah, it seems that `ale#other_source#ShowResults` should be called for each source even when it's empty, according to #1296.

chemzqm:
The problem is you wouldn't have source name when such kind of diagnostic item doesn't exists any more.

yous:
Is there a way to get available/enabled sources for each collection?

chemzqm:
coc-python could add source name to diagnostic message.

Issue #1054, opened 2019-08-01
Add support for vim's new `tagfunc` method


jbbudzon:
**Is your feature request related to a problem? Please describe.**
When I used tags, I would often leverage the tagstack as opposed to the jumplist. An example would be using `<C-]>` to jump to a definition, then using a variety of `[[`/`]]` commands to jump around, and then when I am finished I can press `<C-t>` once to get back to where I was before the jump to definition call. This does not work if go-to-definition doesn't push onto the tagstack (currently, the jumplist is used instead).

**Describe the solution you'd like**
Rather than implement custom tagstack pushing/popping functionality, we should leverage vim's new tagfunc method (vim/vim#4010). I feel like this hooks into vim in a very native way, similar to #181 (which I ❤️, such a great change!!). By implementing a tagfunc, we will hook into all tag functionality and automatically pick up mappings like `<C-w>]`.

**Describe alternatives you've considered**
I've considered using tools that manipulate the tagstack for you. But these tools/plugins are old and do not implement tagfunc. This means every tag mapping/command/etc needs to be overwritten to use the tool/plugin to manage the stack and then finally calling coc.nvim to make the jump. I'm also trying to dig down into coc.nvim code and see how I may go about implementing something like this on my own.

**Additional context**
Finding native ways to hook into vim is always a good thing. Similar to #181, I feel that this enhancement will improve the coc.nvim experience.

Ende93:
Thank you, but @chemzqm is too busy to do it recently, pull request is welcome.

chemzqm:
We can add tag support for jumps, could be great if someone can make a PR.

jbbudzon:
```viml
function! coctagfunc#tagfunc(pattern, flags, info) abort
	if a:flags != "c"
		return v:null
	endif

	let name = expand("<cword>")
	execute("call CocAction('jumpDefinition')")
	let filename = expand('%:p')
	let cursor_pos = getpos(".")
	execute("normal \<C-o>")
	return [ { 'name': name, 'filename': filename, 'cmd': string(cursor_pos[1]) } ]
endfunction
```

This was the TERRIBLY crude POC I was thinking of an it does seem to work! Preferably, I'd find a way to get the `filename` and `cursor_pos` without making the jump and then jumping back.

Also, it would be nice to find out how to handle tag identifiers when no cursor position or context is available. But, I see this as a phase 2. I think the right thing to do here may be to essentially perform the symbols search that CocList provides.

[...]

PR #1380, opened 2019-12-04
Add basic tagfunc support using current location

ms:
(Just a draft for now)

Long time user, first time contributor. I'm fine with JS/TS but only so-so in terms of VimL and I'm not familiar with the codebase at all so I welcome any and all direct feedback.

This attempts to address part of https://github.com/neoclide/coc.nvim/issues/1054

I went the route of putting as much logic in the JS code as possible. With something as simple as:
```viml
function! CocTagFunc(pattern, flags, info)
  return CocAction('getTagFuncLocations', a:pattern, a:flags, a:info)
endfunction
set tagfunc=CocTagFunc
```

 `^]` seemed to work (and `^T`!)

As you can tell, this does not (yet) support arbitrary `:tags` with no cursor location or completion using `^X^]` but I am more than willing to work on that with a few pointers/feedback on the existing approach.


jbbudzon:
This is a much better version of my VimL hack! Now that this is implemented on the node.js side, it should be much easier to hook into the rest of coc.nvim. I was hoping that for handling tagfunc when no cursor location is present, we could just hook into CocList -> Symbols 😉

bakshizaki:
Excited for this feature. Hope it makes to master soon. 😃

jbbudzon:
Curious as to what it would take to get this into upstream. The concept seems sound, although it would be interesting to hear from the maintainer on the general direction of the feature.

chemzqm:
I've added `CocTagFunc` method to coc.nvim.

PR #1425, opened 2019-12-25
Fix stuck when jump somewhere synchronous

timqi:
change action calls below to async

coc-definition
coc-declaration
coc-implementation
coc-type-definition
coc-references


chemzqm:
It's intended to avoid issues on vim8

Avi-D-coder:
@chemzqm could it be async on neovim

chemzqm:
@Avi-D-coder use `CocActionAsync` function instead for keymapping.

Avi-D-coder:
I already am. We could default to async on neovim

chemzqm:
@Avi-D-coder No, I would prefer synchronous location jump to avoid invalid editor state.

PR #1449, opened 2020-01-06
CocCurrentWorkspacePath returns workspace path

dezza:
I think it's valuable to have a function to easily return the current buffer workspace path.

This makes it easier to make mappings to run/build projects by reusing functionality Coc already has.


fannheyward:
I prefer to just add `currentWorkspacePath` method, but not `CocCurrentWorkspacePath` command. We can use `CocAction('currentWorkspacePath')` instead.

dezza:
Alright, I did a CocAction instead.

PR #1450, opened 2020-01-07
feat(workspace): rm .vim from default rootPatterns

fannheyward:
`.vim` folder is always in $HOME, used as default pattern will always resolved `$HOME` to rootPath


fannheyward:
@chemzqm I've merged this first.

PR #1464, opened 2020-01-14
feat(suggest): suggest.asciiCharactersOnly option

### LINKED ISSUE #1459: Provide an option to complete for ASCII-only characters

I was wondering if there is a way to complete when typing ASCII characters, and don't complete for a non-ASCII prefix. If not, could you add an option to achieve this?

Here is the story. I'm a Chinese user. When I'm writing Chinese, the bundled sources (`around`, `buffer` and `file`) will do the completion and show some sentences matching the prefix that I have typed. This is really annoying. Because completing words for Chinese is somewhat unrealistic, I prefer that the coc.nvim could ignore them and don't show these results.

I don't know if there is an existing way to handle this problem. If there is, please let me know.

So my suggestion is to filter the completion results and test whether they contain any non-ASCII characters. Provide an option to configure this behavior, e.g. `completeForAsciiCharactersOnly`.

Or more generally, could there be a mechanism to define some user rules to filter completion results? Just out of curiosity.

fannheyward:
close #1459

**default false**

~cc @chemzqm I make `suggest.asciiCharactersOnly` default true, any suggestions?~

### DIFF
--- a/src/completion/index.ts
+++ b/src/completion/index.ts
@@ -144,6 +144,7 @@ export class Completion implements Disposable {
       localityBonus: getConfig<boolean>('localityBonus', true),
       highPrioritySourceLimit: getConfig<number>('highPrioritySourceLimit', null),
       lowPrioritySourceLimit: getConfig<number>('lowPrioritySourceLimit', null),
+      asciiCharactersOnly: getConfig<boolean>('asciiCharactersOnly', false)
     }
   }
 
--- a/data/schema.json
+++ b/data/schema.json
@@ -464,6 +464,11 @@
       "description": "Invalid character for strip valid word when inserting text of complete item.",
       "default": [" ", "(", "<", "{", "[", "\r", "\n"]
     },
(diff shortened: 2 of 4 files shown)


src/completion/index.ts
```diff
@@ -144,6 +144,7 @@ export class Completion implements Disposable {
       localityBonus: getConfig<boolean>('localityBonus', true),
       highPrioritySourceLimit: getConfig<number>('highPrioritySourceLimit', null),
       lowPrioritySourceLimit: getConfig<number>('lowPrioritySourceLimit', null),
+      asciiCharactersOnly: getConfig<boolean>('asciiCharactersOnly', false)
```
voldikss: The default value is `true` in `schema.json` but  `false` here. could be a forgetten work

fannheyward:
@voldikss bingo, fixed

chemzqm:
@fannheyward to not break current behavior, It should be false by default.

fannheyward:
Make default as false.

PR #1476, opened 2020-01-17
CocAction('workspaceInitialized') returns finished-loading-bool of workspace

dezza:
For extending the reusability of exposing workspacePath here is a command that presents the loading-state of workspace to avoid running commands too early.


chemzqm:
Why you need such kind of API?

dezza:
For mappings specific to building and executing projects (e.g Run-command)

chemzqm:
You should use `g:coc_workspace_initialized` and `CocNvimInit` autocmd instead.

dezza:
alright missed those on checking thank you

PR #1480, opened 2020-01-18
feat(handler): formatOnTypeFiletypes support

### LINKED ISSUE #1472: restrict `coc.preferences.formatOnType` for certain file type only
The function for format file on saving file can be customized for certain file types, but Iit seems that `coc.preferences.formatOnType` can only be toggle by `true` or `false`. Is it possible to expand this setting for file types customization.

fannheyward:
This will only take effect when `coc.preferences.formatOnType` is `true`, and will only check this setting when not empty, in order to avoid breaking change.

close #1472

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -853,6 +853,14 @@
       "description": "Set to true to enable format on type",
       "default": false
     },
+    "coc.preferences.formatOnTypeFiletypes": {
+      "type": "array",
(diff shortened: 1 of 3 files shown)


yech1990:
To make the code more consistent, I think `formatOnInsertLeaveformatFiletypes` and `formatOnSave` are necessary.

PR #1506, opened 2020-01-28
Added option to include the kind in filtering the outline

cosminadrianpopescu:
The outline coc list is not quite that useful (at least for me) if I cannot filter also by kind. 

What I want to do is this:

`CocList --normal --no-sort --input=Method outline -kind` and see only the methods form the current file, which in a lot of cases will be the current class. 

Then I can even press `i` and add a method name in input (for example `Method doHighlight`) and find my method.

The issue for me is that sometimes I don't know the method name, so I want to have a class outline. Hope this clarifies. 

The option by default is false, which means that the behavior is the current one.

### DIFF
--- a/src/list/source/outline.ts
+++ b/src/list/source/outline.ts
@@ -4,7 +4,7 @@ import { URI } from 'vscode-uri'
 import which from 'which'
 import languages from '../../languages'
 import Document from '../../model/document'
-import { ListContext, ListItem } from '../../types'
+import { ListContext, ListItem, ListArgument } from '../../types'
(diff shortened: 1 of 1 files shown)


oblitum:
@cosminadrianpopescu shouldn't this also be listed in `:h coc-list-options`?

cosminadrianpopescu:
> @cosminadrianpopescu shouldn't this also be listed in :h coc-list-options?

Actually, I don't think so. This is a specific option for the outline list. It's not a general option.

PR #1520, opened 2020-02-03
adding number highlighting

danjenson:
Highlights numbers with the same style as the previous signs; this eliminates the need for the signcolumn if you don't want it, i.e. numbers are red for errors and yellow/brown for warnings.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -545,6 +545,11 @@
       "default": true,
       "description": "Enable signs for diagnostics."
     },
+    "diagnostic.enableHiglightLineNumber": {
+      "type": "boolean",
+      "default": true,
+      "description": "Enable higlighting line numbers for diagnostics."
+    },
     "diagnostic.enableMessage": {
       "type": "string",
       "default": "always",
--- a/data/schema.json
+++ b/data/schema.json
@@ -545,10 +545,10 @@
       "default": true,
       "description": "Enable signs for diagnostics."
     },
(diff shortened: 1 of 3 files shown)


oblitum:
* I think this needs to be optional, people may not be expecting their line column to change.
* If this is NeoVim specific, you need to add checks to not break on Vim.

danjenson:
I made it optional. The call should only be valid for `nvim` since it is issued with `nvim.command(...`, so it shouldn't be executing with `vim`. The error looks unrelated (maybe a flaky test?).

oblitum:
> The call should only be valid for nvim since it is issued with nvim.command(..., so it shouldn't be executing with vim.

Ah ok.

>  The error looks unrelated (maybe a flaky test?)

Sure.

Please fix typos, there's some "Higlight".

danjenson:
Typos fixed.

danjenson:
Another unrelated error...

iamcco:
`numhl` is neovm only, it may get error when using vim.

oblitum:
@iamcco https://github.com/neoclide/coc.nvim/pull/1520#issuecomment-581673290

iamcco:
> The call should only be valid for nvim since it is issued with nvim.command(..., so it shouldn't be executing with vim

When using vim, `nvim` is vim instance

oblitum:
Ah OK, so it now needs a fix indeed.

fannheyward:
Fixed

PR #1527, opened 2020-02-05
feature(config): added env variable to change default config folder

chrisands:
Added config env variable that allow change default config path


fannheyward:
You can use `:h g:coc_config_home` to set config path, no need to use env vars.

PR #1548, opened 2020-02-13
feat(diagnostic): checkCurrentLine is a fallback behavior

sam-mccall:
When checkCurrentLine is enabled but there are diagnostics exactly under the
cursor, only show those. If there are none, show all diagnostics from the line.

That way it's still easy to inspect diagnostics with small ranges, but
it's possible to distinguish when there's several diagnostics on a line.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -486,7 +486,7 @@
     },
     "diagnostic.checkCurrentLine": {
       "type": "boolean",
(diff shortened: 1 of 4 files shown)


sam-mccall:
This could also be a third mode, though I suspect it's not needed.

I guess the doc for `checkCurrentLine` in coc.cnx needs to be updated too, but I can't do it.

PR #1559, opened 2020-02-17
feat(completion) support `score` extension on LSP CompletionItem

sam-mccall:
Client-side filtering and ranking of LSP results is great for latency,
but can't take into account server-side signals such as popularity,
type-matches-context etc.

This extension lets the server provide a numeric score multiplier.
It's documented here: https://clangd.github.io/extensions.html
clangd implements it at trunk and in the pending 10.0 release.

This is the main blocker for recommending coc.nvim for clangd. (currently we
recommend YCM and just turn all client-side filtering off).
https://github.com/clangd/clangd/issues/284

My attempt to get something like this standardized doesn't seem to be moving:
https://github.com/microsoft/language-server-protocol/issues/348

### DIFF
--- a/src/__tests__/modules/completion.test.ts
+++ b/src/__tests__/modules/completion.test.ts
@@ -439,6 +439,31 @@ describe('completion TextChangedP', () => {
     expect(items[0].word).toBe('foo#abc')
   })
 
+  it('should use source-provided score', async () => {
+      let source: ISource = {
+          priority: 0,
+          enable: true,
+          name: 'source',
+          sourceType: SourceType.Service,
+          doComplete: (_opt: CompleteOption) : Promise<CompleteResult> => {
+              return Promise.resolve({
+                  items: [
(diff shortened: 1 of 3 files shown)


oblitum:
This is nice. Do you know whether there's any other server that provides it?

sam-mccall:
I don't know of any.
I'd like to think languages more modern than C/C++ avoid dumping thousands of symbols into one namespace so this isn't so pressing :-)

fannheyward:
I've created [coc-clangd](https://github.com/fannheyward/coc-clangd), added some clangd protocol extensions support:

1. Switch between source/header, using `:CocCommand clangd.switchSourceHeader`
1. File status monitor, shows on statusline
1. Force diagnostics generation, default true
1. Diagnostic categories & inline fixes, can be used for diagnostic format
1. Symbol info under cursor, using `CocCommand clangd.symbolInfo`, currently just echoed

PR #1564, opened 2020-02-19
feat(diagnostics): show diagnostic message first, hide severity

sam-mccall:
This is aimed at improving the clarity/readability of the popups.

There's currently a lot of "tags" displayed before the diagnostic
message in the floating window.

The message is usually the most relevant information, but starts partway
through the line and is often wrapped or truncated. Having to hunt for
this info conflicts with the immediate/contextual nature of the popup.

The severity is almost always redundant: the text is colored and the
gutter signs show the severity too.

Before:
```
[clang ovl_no_viable_function_in_call] [E] No matching function for call to 'foo'
```
image
image

After:
```
No matching function for call to 'foo'
[clang: ovl_no_viable_function_in_call]
```
image
image

(Personally I'd probably prefer to drop the source information from the
hover version entirely, and keep it for `diagnosticPreview` where
there's more space. But that's a more invasive change!)

### DIFF
--- a/src/__tests__/modules/diagnosticManager.test.ts
+++ b/src/__tests__/modules/diagnosticManager.test.ts
@@ -200,7 +200,7 @@ describe('diagnostic manager', () => {
     let bufnr = await nvim.call('nvim_win_get_buf', winid) as number
     let buf = nvim.createBuffer(bufnr)
     let lines = await buf.lines
-    expect(lines.length).toBe(3)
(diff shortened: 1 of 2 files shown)


fannheyward:
I've added `diagnostic.format`, https://github.com/neoclide/coc.nvim/commit/839f964fe9b0c44a9857c63d31d1687bc0b9a2aa you can define your own format.

PR #1565, opened 2020-02-19
fix(completion): support allCommitCharacters as fallback for commitCharacters

sam-mccall:
LSP says if commitCharacters is not provided for a particular completion option,
CompletionOptions.allCommitCharacters should be used instead.


sam-mccall:
Tested with clangd head.

I'm not sure the best way/whether to write a test for this, the existing commitCharacters tests don't test the LSP side, they mock out shouldCommit.

PR #1589, opened 2020-02-27
fix(completion): add sourceScore for CompletionItem

fannheyward:
PR #1559 adds score from LS to CompletionItem, but coc will compute `item.score` as user types, so we can't use `item.score` to save score from LS.

### DIFF
--- a/src/__tests__/modules/completion.test.ts
+++ b/src/__tests__/modules/completion.test.ts
@@ -448,8 +448,8 @@ describe('completion TextChangedP', () => {
           doComplete: (_opt: CompleteOption) : Promise<CompleteResult> => {
               return Promise.resolve({
                   items: [
-                      { word: 'candidate_a', score: 0.1 },
-                      { word: 'candidate_b', score: 10 },
+                      { word: 'candidate_a', sourceScore: 0.1 },
+                      { word: 'candidate_b', sourceScore: 10 },
                       { word: 'candidate_c' },
                   ]
               })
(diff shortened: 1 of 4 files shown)


chemzqm:
```
item.score = input.length ? score * (item.sourceScore || 1) : 0
```
We should not change `score` which is used for sort at first, it's something like `sortText` of complete item, I will fix it.

sam-mccall:
@chemzqm I fear we're going in circles a bit here, sorry for the confusion/bad implementation in the first patch.

[The `score` extension](https://clangd.github.io/extensions.html#code-completion-scores) is AFAIK only implemented by clangd, which also sets a sortText consistent with it, so I think it's now a no-op after neoclide/coc.nvim@712ead3

The problem the extension is trying to solve: when the server uses sortText to communicate ranking, information is lost. When applying client-side filtering/fuzzy-matching, server-side signals (such as #refs to the candidate item) aren't known to the client. The client can either:
1. sort by (sortText, nameMatchScore), effectively just filtering the server-provided items
2. sort by (nameMatchScore, sortText), effectively re-ranking all items ignoring server-side signals

coc.nvim basically does 2, as I understand.

The idea with providing numeric score is that the client can sort by nameMatchScore * serverScore. This is what 3c462c6 attempted to do, and I think 1473e00 did better.

chemzqm:
@sam-mccall thanks for the link, I will revert my change.

chemzqm:
Fixed.

sam-mccall:
@chemzqm thanks! Happy to help if this causes problems in future

PR #1602, opened 2020-03-04
fix(outputChannel): don't wrap in log files

sam-mccall:
Not sure whether PRs for settings like this are appreciated, but this is driving me a little nuts, I don't think it's the right default.

Output windows open in a vsplit (so often narrow) and having logs wrap is annoying:
 - it hides the typical repetetive structure that makes logs easy to
   visually scan
 - in the presence of very long lines (which are reasonably common) you
   often see only a couple of log entries, and half the screen is blank

Current:
wrap
(Note half the window is displaying nothing, though there are more logs, because the next line is very long)

Patched:
nowrap


sam-mccall:
(cc @kadircet who was also bothered by this I think)

PR #1603, opened 2020-03-04
fix(float): stop floats from getting too small on the RHS of the screen.

sam-mccall:
Currently if a float for diagnostics always triggers to the right
of the cursor. This is fine when the cursor is near column 0, but the right edge
of the window limits the size, so if the cursor is at say 70/80, there
are only 10 columns available and the content wraps *a lot*.

Compounding this, maxheight is set to number of lines of content, but
isn't increased if the content has to wrap. So when the content
wraps, it's truncated. (Scrollable, but this is awkward)

Current:
old_000

Patched:
new_004

Looking at the history, it seems there's been some tension between
wrapping and fixing widths. 77cbc92ca5c38e1be moved the window to the
left if there wasn't enough room (so it wouldn't wrap), but then
46e0f4d8e5d90ed42 made it conditional on offsetX != 0, which meant
everything except signature help wraps again.

So the proposal here is:
 - on the LHS things stay as they are
 - as the cursor moves to the right, the popup eventually hits the right
   edge, it starts to shrink+wrap
 - once it reaches a minimum width (50 cols, could be a setting), it
   stops shrinking and is no longer locked to the cursor position,
   rather to the RHS of the window.
 - the popup is allowed to grow vertically to accommodate the content
   when it wraps, up to the global maxheight.
 - popups with offsetX set (signature help) are fixed-width: minimum width is their content width

Graphically, where the cursor is `X`:

```
  X[------------------]          |
       X[------------------]     |
            X[------------------]|
                 X[-------------]|
                  [----X--------]|
                  [--------X----]|
                  [------------X]|
```

### DIFF
--- a/src/model/floatFactory.ts
+++ b/src/model/floatFactory.ts
@@ -138,11 +138,13 @@ export default class FloatFactory implements Disposable {
     if (alignTop) docs.reverse()
     await this.floatBuffer.setDocuments(docs, maxWidth)
     let { width } = this.floatBuffer
-    if (offsetX) {
-      offsetX = Math.min(col - 1, offsetX)
-      if (col - offsetX + width > columns) {
-        offsetX = col - offsetX + width - columns
-      }
+    // Ensure the floating window isn't tiny if the cursor is on the right:
+    // increase the offset to accommodate some minimum width.
+    // If we have offsetX, precise positioning is intended, force exact width.
+    let minWidth = offsetX ? width : Math.min(width, 50, maxWidth)
+    offsetX = Math.min(col - 1, offsetX)
+    if (col - offsetX + minWidth > columns) {
+      offsetX = col - offsetX + minWidth - columns
     }
     this.alignTop = alignTop
     return {
@@ -216,7 +218,7 @@ export default class FloatFactory implements Disposable {
         minwidth: config.width - 2,
         minheight: config.height,
         maxwidth: config.width - 2,
-        maxheight: config.height,
+        maxheight: this.maxHeight,
         firstline: alignTop ? -1 : 1
       })
       this.floatBuffer.setLines()


cristobaltapia:
Hi, I see that this was merged. However, I don't see the expected behavior. My popup keeps getting narrower as the cursor approaches to the RHS. Is there a configuration I have to change? Is this only for error messages or should it work with the completion menu too?

chemzqm:
@cristobaltapia Unable to reproduce what you have describe, your coc.nvim maybe too old.

cristobaltapia:
That might be. This is what I see when using coc-texlab:
Screenshot-2020-05-22-160029
Screenshot-2020-05-22-160106

And this is my config:
```json
{
    "diagnostic.displayByAle": false,
    "suggest.labelMaxLength": 25,
    "suggest.detailMaxLength": 100,
    "snippets.ultisnips.directories": [
        "UltiSnips",
        "plugged/MySnippets/Ultisnips",
        "~/Templates/ultisnips-templates"
    ],
    // Python
    "python.formatting.provider": "yapf",
    "markdownlint.config": {
        "default": true,
        "MD013": false,
        "MD003": false
    },
    "python.formatting.yapfPath": "yapf",
    "python.formatting.blackPath": "black",
    "python.linting.enabled": true,
    "python.linting.pylintEnabled": true,
    "python.linting.flake8Enabled": false,
    "python.linting.mypyEnabled": false,
    "python.linting.prospectorEnabled": false,
    "python.linting.pydocstyleEnabled": true,
    "python.linting.pydocstyleArgs": [
        "--convention=numpy"
    ],
    "python.linting.lintOnSave": true,
    "python.linting.maxNumberOfProblems": 200,
    "python.terminal.executeInFileDir": true,
    "python.venvFolders": [
        ".virtualenvs",
        "venv"
    ],
    "python.venvPath": "/home/tapia/.virtualenvs",
    "python.jediEnabled": false,
    "suggest.enablePreview": true,
    "suggest.minTriggerInputLength": 2,
    "suggest.fixInsertedWord": true,
    "suggest.autoTrigger": "always",
    "coc.source.file.enable": true,
    "coc.source.vimtex.enable": false,
    "coc.source.around.enable": true,
    "latex.lint.onChange": false,
    "latex.build.onSave": false,
    "yaml.format.enable": true,
    "yaml.format.bracketSpacing": true,
    "yaml.completion": true,
    "yaml.hover": true,
    "coc.source.ultisnips.enable": true,
    "coc.source.ultisnips.filetypes": [
        "python",
        "tex",
        "julia",
        "bash",
        "markdown",
        "markdown.pandoc",
        "css",
        "bib",
        "make",
        "javascript",
        "vim",
        "snippets"
    ],
    "highlight.colors.enable": true,
    "highlight.document.enable": false,
    "git.addedSign.hlGroup": "GitGutterAdd",
    "git.changedSign.hlGroup": "GitGutterChange",
    "git.removedSign.hlGroup": "GitGutterDelete",
    "git.topRemovedSign.hlGroup": "GitGutterDelete",
    "git.changeRemovedSign.hlGroup": "GitGutterChangeDelete",
    "session.directory": "~/.vim/sessions",
    "yank.highlight.enable": false,
    "yank.enableCompletion": false,
    "languageserver": {
        "julia": {
            "command": "/usr/bin/julia",
            "args": [
                "--startup-file=no",
                "--history-file=no",
                "-e",
                "using LanguageServer; using Pkg; import StaticLint; import SymbolServer; env_path = dirname(Pkg.Types.Context().env.project_file); debug = false; server = LanguageServer.LanguageServerInstance(stdin, stdout, debug, env_path, \"\", Dict()); server.runlinter = true; run(server);"
            ],
            "filetypes": [
                "julia"
            ]
        },
        "dockerfile": {
            "command": "docker-langserver",
            "filetypes": [
                "dockerfile"
            ],
            "args": [
                "--stdio"
            ]
        },
        "bash": {
            "command": "bash-language-server",
            "args": [
                "start"
            ],
            "filetypes": [
                "sh",
                "make"
            ],
            "ignoredRootPaths": [
                "~"
            ]
        },
        "ccls": {
            "command": "ccls",
            "filetypes": [
                "c",
                "cpp",
                "objc",
                "objcpp"
            ],
            "rootPatterns": [
                ".ccls",
                "compile_commands.json",
                ".vim/",
                ".git/",
                ".hg/"
            ],
            "initializationOptions": {
                "cache": {
                    "directory": "/tmp/ccls"
                }
            }
        },
        "efm": {
            "command": "efm-langserver",
            "args": [],
            // custom config path
            // "args": ["-c", "/path/to/your/config.yaml"],
            "filetypes": [
                "eruby"
            ]
        },
        "vimls": {
            "module": "/usr/lib/vim-language-server/out/index.js",
            "args": [
                "--node-ipc"
            ],
            "initializationOptions": {
                "iskeyword": "@,48-57,_,192-255,-#", // vim iskeyword option
                "vimruntime": "", // $VIMRUNTIME option
                "runtimepath": "", // vim runtime path separate by `,`
                "diagnostic": {
                    "enable": true
                },
                "indexes": {
                    "runtimepath": true, // if index runtimepath's vim files this will effect the suggest
                    "gap": 100, // index time gap between next file
                    "count": 3, // count of files index at the same time
                    "projectRootPatterns": [
                        "strange-root-pattern",
                        ".git",
                        "autoload",
                        "plugin"
                    ] // Names of files used as the mark of project root. If empty, the default value [".git", "autoload", "plugin"] will be used
                },
                "suggest": {
                    "fromVimruntime": true, // completionItems from vimruntime's vim files
                    "fromRuntimepath": false // completionItems from runtimepath's vim files, if this is true that fromVimruntime is true
                }
            },
            "filetypes": [
                "vim"
            ]
        }
    }
}

```

chemzqm:
It's expected since it's already using max width instead of narrow down your content

cristobaltapia:
Ok, thanks. However, I am still a bit confused. Is there a way to change that? because I am missing important information, that is not displayed there.

PR #1608, opened 2020-03-04
increase max timeout to 15000

daern91:
I've currently experienced some issues with `rust-analyzer` taking longer than default 5s, and I seem to not be the only one. Reference below,
https://gitter.im/neoclide/coc.nvim?at=5e0acd6c833c373f4d84eae4

PR increases the max timeout to 15s without touching default. PR also sets the correct minimum value.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -376,7 +376,7 @@
       "type": "integer",
       "default": 5000,
       "minimum": 500,
-      "maximum": 5000,
+      "maximum": 15000,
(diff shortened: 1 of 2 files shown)


chemzqm:
I would consider the server is useless when it doesn't return results after 5s.

daern91:
Good point. The main problem is indeed with the server but I also think this helps with debugging. 

Thanks a lot for the quick merger!

PR #1613, opened 2020-03-05
fix #1601: check placeholder when accepting completion

### LINKED ISSUE #1601: gopls completion within snippet cancels the snippet session
This issue is moved from https://github.com/neoclide/coc-snippets/issues/101

coc-snippet-error

Steps to reproduce:

1. minimal-coc.nvim
```
set nocompatible
set runtimepath^=/home/jackieli/.config/nvim/plugged/coc.nvim
filetype plugin indent on
syntax on
set hidden
```
2. `CocConfig`
```
{
  "languageserver": {
    "golang": {
      "command": "gopls",
      "initializationOptions": {
        "usePlaceholders": false,
        "completeUnimported": true,
        "static_check": true,
        "hoverKind": "FullDocumentation"
      },
      "rootPatterns": ["go.mod", ".vim/", ".git/", ".hg/"],
      "filetypes": ["go"]
    }
  }
}
```
3. `nvim -u ~/.config/nvim/minimal-coc.vim main.go`
4. `:CocCommand snippets.editSnippets` and insert a simple snippet
```
snippet fori "for 0..N-1 { ... }"
for ${1:i} := 0; $1 < ${2:N}; $1++ {
	${0:${VISUAL}}
}
endsnippet
```
5. edit main.go like so:
```
package main

func main(){
   fori<C-n><C-y> // C-n to go to the snippet, C-y to accept the snippet
}
```
6. use `<C-j>` to jump to the second placeholder, and input flag.N, and then `<C-y>` to accept flag.Narg(), then press `<C-j>` to jump to the next placeholder, at this point, line break is inserted instead of jumping to the next placeholder

Further diagnostic shows that in [`snippets/session.ts:136`](https://github.com/neoclide/coc.nvim/blob/master/src/snippets/session.ts#L136). the `edit.range` returns the wrong range: edit.range.start.line should be 16, but it's 15. I suspect this has something to do with: https://github.com/neoclide/coc.nvim/issues/1001

If comment out the `this.deactivate` within `if (!placeholder)` check block, it works

jackielii:
This is my very naive attempt to fix the problem. It works, but far from ideal. It's relatively small, so please point out the proper way to fix this

### DIFF
--- a/src/snippets/session.ts
+++ b/src/snippets/session.ts
@@ -134,9 +134,19 @@ export class SnippetSession {
     }
     let placeholder = this.findPlaceholder(edit.range)
     if (!placeholder) {
-      logger.info('Change outside placeholder, cancelling snippet session')
-      this.deactivate()
-      return
+      // issue #1601
+      // if we accept a completion, then the edit.range.end will be bigger than
+      // placeholder.range.end, because the placeholder.range.end is constantly
+      // updated when we're typeing in the same placeholder. However we're
+      // still in a right placeholder. In this case, we just want to use the
+      // current placeholder
+      if (comparePosition(edit.range.start, this.placeholder.range.start) === 0) {
+        placeholder = this.placeholder
+      } else {
+        logger.info('Change outside placeholder, cancelling snippet session')
+        this.deactivate()
+        return
+      }
     }
     if (placeholder.isFinalTabstop) {
       logger.info('Change final placeholder, cancelling snippet session')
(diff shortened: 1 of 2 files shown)


chemzqm:
No, the fix is wrong, the change range must be contained.

jackielii:
@chemzqm thanks for reviewing this. I did say this is a very naive approach to fix the problem. It does highlight the problem though: when accepting the completion, the `edit.range.end` is bigger than `placeholder.range.end`.

Please point me at the right direction and I'll try to do the proper fix.

Thanks & Regards,

chemzqm:
The `edit.range.end` should not bigger than `placeholder.range.end`, we need to figure out why this happened.

PR #1620, opened 2020-03-05
don't force edit in snippet

### LINKED ISSUE #1601: gopls completion within snippet cancels the snippet session

jackielii:
If forced, the edit will trigger the snippet update logic, and the range check will fail, resulting in snippet session getting cancelled

fixes #1601

### DIFF
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -677,7 +677,7 @@ class Languages {
       await doc.applyEdits(nvim, [{
         range: Range.create(linenr - 1, 0, linenr, 0),
         newText: `${start}${end}\n`
-      }])
+      }], false)
       // can't select, since additionalTextEdits would break selection
       let pos = Position.create(linenr - 1, range.start.character)
       return await snippetManager.insertSnippet(newText, false, Range.create(pos, pos))


oblitum:
As commented previously, this patch brings more issues, it causes nested snippet expansion to not jump correctly to first parameter.

jackielii:
@oblitum you're right. I'm able to reproduce in my case as well. I reckon there needs to be some special logic to handle nested snippets

oblitum:
Ask for reopen if there's anything new as fix.

PR #1637, opened 2020-03-08
show error message when extension's main file doesn't exist.

jummy233:
This is a response to this [issue](https://github.com/neoclide/coc-tsserver/issues/125) according to this [problem](https://github.com/neoclide/coc-tsserver/issues/126).

Currently when the main file in package.json doesn't exist extension module will only create an ExtensionExport with empty activate function. But when you download extensions from npm the package itself might have a problem, and if that happened there is no explicit error message.

#### Current createExtension will return a `ExtensionExport` without showing any messages.
2020-03-08-055139_1134x187_scrot

2020-03-08-055316_1078x144_scrot

The pull request shows an error message for this purpose without change the original implementation of `createExtension`

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -764,18 +764,21 @@ export class Extensions {
         }
         isActive = true
         if (!ext) {
+          // check if activate is empty function like () => {}
+          // keep the implementation of createExtension
+          const stringifyFunction = f => f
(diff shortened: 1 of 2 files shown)


chemzqm:
We just need to check if main file exits.

PR #1643, opened 2020-03-10
Fix the discrepancy between maxDetailLength's documented and actual value

kirillbobyrev:
maxDetailLength is documented to be 100 by default but in the
code it appears to be 50. Documentation is updated to reflect current
behaviour.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -298,12 +298,12 @@
     "suggest.labelMaxLength": {
       "type": "number",
       "description": "Max length of abbr that shown as label of complete item.",
-      "default": 200
+      "default": 50
     },
     "suggest.detailMaxLength": {
       "type": "number",
       "description": "Max length of detail that should be shown in popup menu.",
-      "default": 100
+      "default": 50
     },
     "suggest.detailField": {
       "type": "string",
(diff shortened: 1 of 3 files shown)


chemzqm:
I don't want to change that.

kirillbobyrev:
Okay, I reduced the patch to fix the existing discrepancy only.

The problem for us is that having `maxLabelLength` that long leads to the UI that makes it hard to use code completion and we would like to trim it down by default. Is there any way we could easily do that in the coc-clangd extension?

fannheyward:
You should not change the default value, that's a breaking change. You can set `suggest.detailMaxLength` in your coc-settings.json to overwrite the default value.

kirillbobyrev:
"The default value" is inconsistent and has different values in schema and in the [actual code](https://github.com/neoclide/coc.nvim/blob/master/src/languages.ts#L172) (by the way, `detailField` is also set to `abbr` by default there while the documentation mentions `menu` as the default value). What I'm doing is making sure it's the same value in two places.

Here is your [commit](https://github.com/neoclide/coc.nvim/commit/f3a41b7522e587b68260f27027a08510aa80d54f) doing essentially the same. Am I missing something?

fannheyward:
My commit https://github.com/neoclide/coc.nvim/commit/f3a41b7522e587b68260f27027a08510aa80d54f just correct the code value to schema value. We should consistent with scheme defines.

PR #1648, opened 2020-03-11
Improve formatting of workDoneStatus message

expipiplus1:
Only show the integral part of the percentage, does anyone need the
loading status to 1/10000th precision?

Move the percentage marker to the beginning of the message; this way it
stays in the same position as the length of the title and message
changes (assiming right-justified text)


fannheyward:
Changed on master directly, because this PR's code format is broken with master.

expipiplus1:
Thanks!

PR #1656, opened 2020-03-12
Filter the items added to the list history

cosminadrianpopescu:
Normally, when closing a list without selecting an item, I would like for the item to not be added to the history. I should be able to add an item to the history only if an action has been performed on the item. 

For this purpose I added the option `historyAddOnClose`, which is default `true`. This means that it's perfectly backwards compatible. Nothing changes unless this option is set to `false`. If this option is false, then an item will be added to the lists history only if an action has been performed.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -764,6 +764,11 @@
       "default": {},
       "description": "Custom keymappings on insert mode."
     },
+    "list.historyAddOnClose": {
+      "type": "boolean",
+      "default": true,
+      "description": "If set to false, an item will be added to the prompt history only when an action is actually executed"
+    },
     "list.source.outline.ctagsFilestypes": {
       "type": "array",
       "default": [],
--- a/src/list/manager.ts
+++ b/src/list/manager.ts
@@ -39,6 +39,7 @@ export class ListManager implements Disposable {
   public config: ListConfiguration
   public worker: Worker
   private plugTs = 0
+  private historyAddOnClose = false
   private disposables: Disposable[] = []
   private savedHeight: number
   private args: string[] = []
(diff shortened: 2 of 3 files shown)


chemzqm:
No, it's wired since this would add prompt history for actions that cancel the list.

cosminadrianpopescu:
No, this is exactly what happens now. So, I open a list with MRU files. I click up <ctrl-p> and I go to the first item which is file 1. I click again 2 times <ctrl-p>, the item history is now file3.ts. Now, I click ok and file3.ts is opened. 

Then I open again the MRU list. I click <ctrl-p> in history and the first item is file3.ts (which is fine). But now I click again <ctrl-p> and the item is file1.ts. Now, if I click esc (I cancel the list), and then open the MRU list again and again click <ctrl-p>, what I expect there to find is the last opened file (which would be file3.ts), but instead I find file1.ts, which is wrong in my opinion. 

And if this is a desired behavior, this is why I put an option with the default option to keep this behavior.

chemzqm:
@cosminadrianpopescu I can understand now, we should not save input when it's created from history, but your approach is wrong.

cosminadrianpopescu:
Can you elaborate what it's wrong? I am willing to change the PR code. The current approach was built so that the current behavior is not touched. 

The issue is that the current behavior is quite annoying (at least for me). I've also consulted 2 UX designers which confirmed that it's a weird behavior. I don't want to save everything in history just because I looped through it.

chemzqm:
I've made the fix with b7a597f7

cosminadrianpopescu:
Thank you.

PR #1668, opened 2020-03-18
fix(util): suppress 'press ENTER' in echo_messages()

sam-mccall:
This is used for workspace/showMessage, when it contains multiple lines.
Using `echomsg` in a loop means that each line after the second prompts
before displaying.

The downside of this change is if there are multiple consecutive calls
to `echo_messages()`, only the last one is shown. (All will appear in :messages)

Motivating use: clangd has "dump AST" code action behind the
-hidden-features flag. This action calls showMessage with a payload like:
```
FunctionDecl 0x14387ee101e8 </home/sammccall/test.cc:1:1, line:3:1> line:1:6 foo 'void ()'
`-CompoundStmt 0x14387ee10380 <col:11, line:3:1>
  `-DeclStmt 0x14387ee10368 <line:2:3, col:10>
      `-VarDecl 0x14387ee10300 <col:3, col:7> col:7 bar 'int'
```

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -270,9 +270,12 @@ function! coc#util#echo_messages(hl, msgs)
   endif
   execute 'echohl '.a:hl
   let msgs = filter(copy(a:msgs), '!empty(v:val)')
+  " :echomsg can't print multiple lines without 'press ENTER' interruptions.
+  " Send the lines to messages first, then display them with :echo.
   for msg in msgs
-    echom msg
+    echom msg | redraw
   endfor
+  echo join(msgs, "\n")
   echohl None
 endfunction
 


chemzqm:
I can't reproduce 'press ENTER' behavior, we use `:echom` to make sure user can get it from `:messages`.

sam-mccall:
> I can't reproduce 'press ENTER' behavior

@chemzqm What did you try? Running `:call coc#util#echo_messages(...)` directly won't reproduce it, as vim defers "press ENTER" if the command-line is open.

This should do it though:
`:call timer_start(1000, { tid -> coc#util#show_messages(['foo','bar','baz'])})`

> we use :echom to make sure user can get it from :messages.

This patch still does that, it uses echom to send the message to `:messages` (with no prompt), and additionally echo to print all the lines at once (with a single prompt)

fannheyward:
> `:call timer_start(1000, { tid -> coc#util#echo_messages('', ['foo','bar','baz'])})`

doesn't reproduce `press ENTER`.

sam-mccall:
You're right, sorry was AFK. let me get a proper reproducer.

sam-mccall:
OK, so far the only way I've been able to reproduce this is with an actual language server sending a multiline `workspace/showMessage`.

e.g. with `coc-clangd`:
 - add `"clangd": { "args": ["-hidden-features"] }` to coc-settiongs.json
 - create a file with `void foo() { int bar; }`
 - at start-of-file, `:call CocActionAsync('codeAction','')`
 - select "Dump FunctionDecl AST"

The patch definitely fixes this for vim 8.1.2269.

It seems I should be able to create the same effect by calling `echo_messages` from either an autocmd or a timer, but I can't and I'm not sure why.

fannheyward:
@sam-mccall no `press ENTER` in nvim with your test case.

sam-mccall:
@fannheyward Indeed, I got nvim set up and verified it's broken in a different way than vim :-)

| | master | with this patch |
|-|-|-|
|vim 8.1.2269 | 'press ENTER' after each line | shows all lines at once|
|nvim v0.5.0-418-g87d892afa | displays only the last line | shows all lines at once|

fannheyward:
Reproduced, I prefer to add this patch.

@chemzqm in this case, `workspace.showMessage('aa\nbb\ncc\ndd')` will only show `dd` on master, with this patch, will show the full message.

sam-mccall:
Thanks!

PR #1679, opened 2020-03-23
added mdx filetype identifier

onichandame:
# Problem

As I tried to transplant a [vscode plugin](https://github.com/mdx-js/vscode-mdx), activationEvents does not trigger on opening mdx files.

# Cause

mdx is not a built-in filetype of vscode language server at the moment.

The [contributes.languages](https://code.visualstudio.com/api/references/contribution-points#contributes.languages) feature of allows vscode extensions to define their own filetype, i.e. extensions mapped to language identifier. Therefore vscode-mdx works fine without the official support of mdx filetype from vscode.

However coc.nvim is lack of this feature at the moment.

# Fix

I will eventually try to implement this feature. But I wish to focus on the original problem, which is to transplant vscode-mdx to coc. So a dirty and temporary workaround is made.

The change has been kept minimal. Please check if anything will be broken by this PR.

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -550,7 +550,7 @@ export class Extensions {
     this.list.push({ id, extension, deactivate, isLocal: true })
     let { contributes } = packageJSON
     if (contributes) {
-      let { configuration } = contributes
+      let { configuration, languages } = contributes
       if (configuration && configuration.properties) {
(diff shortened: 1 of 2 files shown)


chemzqm:
Use `g:coc_filetype_map` instead.

PR #1688, opened 2020-03-26
refactor(diagnostic): clean up DiagnosticCollection.set()

Ferdi265:
This PR cleans up the implementation of `DiagnosticCollection.set()` in `src/diagnostic/collection.ts`.

This eliminates the recursive call of the `(uri: string, diagnostics: Diagnostic[] | null)` overload when originally calling using the `(entries: [string, Diagnostic[] | null][])` overload.

The main change is wrapping the `uri`, `diagnostics` pair into an array `[[uri, diagnostics]]` and then continuing on as if the `entries` overload was called.

This change reduces the complexity of the function and eliminates some nesting and null-checking since it can be done up-front when processing the `entries` array.

### DIFF
--- a/src/diagnostic/collection.ts
+++ b/src/diagnostic/collection.ts
@@ -22,28 +22,31 @@ export default class Collection implements DiagnosticCollection {
   public set(uri: string, diagnostics: Diagnostic[] | null): void
   public set(entries: [string, Diagnostic[] | null][]): void
   public set(entries: [string, Diagnostic[] | null][] | string, diagnostics?: Diagnostic[]): void {
-    if (Array.isArray(entries)) {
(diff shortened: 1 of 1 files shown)


Ferdi265:
forgot to run tests.

some things still don't match up with the previous implementation. currently fixing that.

Ferdi265:
relevant tests pass locally now, should be equivalent to the old version in functionality

Ferdi265:
okay, the failing test (`src/__tests__/list/mappings.test.ts`) is something that doesn't seem remotely relevant to my change, and it even fails on current master.

At least locally, this PR doesn't seem to have introduced any test failures

Ferdi265:
rebased onto latest master, test situation didn't change

PR #1700, opened 2020-03-30
feat: Added contributes.languages support

onichandame:
# Feature Request

vscode provides [a handy method](https://code.visualstudio.com/api/references/contribution-points#contributes.languages) for plugins to work with a wide variety of languages.
The key of this method is to allow plugins to define the languages they want to cope with and the corresponding completion rules declaratively. This feature is not yet implemented in coc.

2 sub-features are required in order to implement this feature in coc.

1. auto-activation of extensions based on file extensions.
2. setup grammar rules declaratively like [this](https://code.visualstudio.com/api/language-extensions/language-configuration-guide)

# Background

Based on the reply from @chemzqm  in [this](https://gitter.im/neoclide/coc-cn?at=5e7870905a64806d5ee48d04), `contributes.languages` has yet been supported and no plan to add it has been made.

# Change

This PR only implements the first subfeature, which is to activate the extension based on the declaration of `contributes.languages` of the extension itself.

# Known Problems

Currently, only file extensions are included. However, according to [the definition](https://code.visualstudio.com/api/references/contribution-points#contributes.languages), filenames and firstLine should also be supported.

I haven't dived into the second subfeature yet as it can be way more complex than the first one.

# Roadmap

1. make sure that this feature is desired and the way to implement it is acceptable
2. add filenames and firstLine support
3. implement the second subfeature

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -28,6 +28,7 @@ import workspace from './workspace'
 const createLogger = require('./util/logger')
 const logger = createLogger('extensions')
 
+logger.debug('test')('hi')
 export type API = { [index: string]: any } | void | null | undefined
 
 export interface PropertyScheme {
@@ -126,6 +127,7 @@ export class Extensions {
     if (global.hasOwnProperty('__TEST__')) return
     for (let item of this.list) {
       let { id, packageJSON } = item.extension
+      this.setupCustomLanguages(packageJSON)
       this.setupActiveEvents(id, packageJSON)
     }
     // check extensions need watch & install
@@ -562,6 +564,7 @@ export class Extensions {
       }
     }
     this._onDidLoadExtension.fire(extension)
+    this.setupCustomLanguages(packageJSON)
     this.setupActiveEvents(id, packageJSON)
   }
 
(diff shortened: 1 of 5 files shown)


chemzqm:
No, you should use viml to setup your language.
This kind of implementation is quite limited.

onichandame:
> No, you should use viml to setup your language.
> This kind of implementation is quite limited.

Could you point me to an example where viml is used to define a custom language? ideally a coc extension would be appreciated.

iamcco:
Maybe what you want  https://github.com/dart-lang/dart-vim-plugin/blob/master/ftdetect/dart.vim

onichandame:
> Maybe what you want https://github.com/dart-lang/dart-vim-plugin/blob/master/ftdetect/dart.vim

Thx but not exactly what I am looking for. So here is my problem:

Many new frameworks or languages have their own file types and grammar rules and vscode extensions have virtually addressed them all. For example [apollo](https://github.com/prisma-labs/vscode-graphql) or  [mdx](https://github.com/mdx-js/vscode-mdx).

The quickest way to implement these extensions into neovim is to transplant them as coc extensions. Based on my effort to transplant these plugins I have found that the `contributes.languages` is one of the key features they require.

If it is not possible to copy-paste from vscode extensions, I will still try to make pure vim plugins that solve my problem, but with a lot more effort and time.

chemzqm:
For filetype support, it's just single line in vim, like:
``` vim
autocmd BufNewFile,BufRead *.tsx setlocal filetype=typescript.tsx
```
or use `g:coc_filetype_map` to fix filetype for coc.nvim.

onichandame:
AFAIK, coc extensions installed by `CocInstall` do not enter the vim plugin lifecycle. So in order to make vimL works in the plugins, they need to be installed directly to vim but not coc, such as using Vim-Plug.

So a simple paradox arises, if the extension works as a sibling of coc but a child of coc, it better not depend on coc. Hence a pure vim plugin is the best option in this case. But as I said, I wish to use coc to manage all the language servers.

If it is not acceptable to support `contributes.languages` as vscode does, I will seek pure vim plugins to solve my problems. Thank you guys for taking time to discuss.

chemzqm:
> If it is not acceptable to support contributes.languages as vscode does

It's not suitable since what you did is only filetype detect in coc.nvim while features including syntax support won't work

PR #1739, opened 2020-04-10
feat(model): clear output when larger than MAX_STRING_LENGTH

fannheyward:
https://github.com/fannheyward/coc-rust-analyzer/issues/98

### DIFF
--- a/src/model/outputChannel.ts
+++ b/src/model/outputChannel.ts
@@ -41,8 +41,8 @@ export default class BufferChannel implements OutputChannel {
   }
 
   public append(value: string): void {
-    if (this._content.length > constants.MAX_STRING_LENGTH) {
-      this.clear()
+    if (this._content.length >= constants.MAX_STRING_LENGTH) {
+      this.clear(10)
     }
     this._content += value
     this.promise = this.promise.then(() => {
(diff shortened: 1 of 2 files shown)


chemzqm:
We should not just clear, but use latest lines  as much as possible.

src/model/outputChannel.ts
```diff
@@ -41,8 +41,8 @@ export default class BufferChannel implements OutputChannel {
   }
 
   public append(value: string): void {
-    if (this._content.length > constants.MAX_STRING_LENGTH) {
-      this.clear()
+    if (this._content.length >= constants.MAX_STRING_LENGTH) {
```
chemzqm, line 44: Should be `this._content.length + value.length`

src/model/outputChannel.ts
```diff
@@ -51,20 +51,25 @@ export default class BufferChannel implements OutputChannel {
   }
 
   public appendLine(value: string): void {
-    if (this._content.length > constants.MAX_STRING_LENGTH) {
-      this.clear()
+    if (this._content.length >= constants.MAX_STRING_LENGTH) {
```
chemzqm, line 54: Should be `this._content.length + value.length`

miuirussia:
```
  src/model/outputChannel.ts:2:10 - error TS2305: Module '"buffer"' has no exported member 'constants'.
  
  2 import { constants } from 'buffer'
             ~~~~~~~~~
  
  
  Found 1 error.
```

fannheyward:
@miuirussia what's your node version?

miuirussia:
12.16.1

fannheyward:
Can't reproduce. constants was added in 8.2.0 https://nodejs.org/api/buffer.html#buffer_buffer_constants, should not cause this error.

chemzqm:
Remove node_modules and run `yarn install` in coc.nvim's folder.

miuirussia:
Maybe in package.json need to declare nodejs version?

PR #1747, opened 2020-04-11
Avoid TypeError on code lens action if there are none available

FallenWarrior2k:


FallenWarrior2k:
So, I just ran the tests on master locally, and the same test fails on master as well, albeit with a slightly different error: there's no timeouts involved and the "Received string" is my current clipboard content instead of an empty string.

oblitum:
Just noticed that this part of the code has some typos, including in error messages. "Lenes" should be "Lenses".

FallenWarrior2k:
I actually fixed that locally but didn't include that in this PR because it didn't fit topically. If desired, I could push that commit or make a second PR.

-------- Original Message --------

PR #1764, opened 2020-04-15
test(list): fix failed paste on ubuntu

fannheyward:
This is a _**quick-dirty**_ fix to make tests passed. From https://github.com/sindresorhus/clipboardy/issues/57#issuecomment-562033847 and my local tests, clipboardy fails on ubuntu and will block tests to run.


fannheyward:
Tests are passed.

PR #1765, opened 2020-04-15
feat(diagnostic): allow disabling the navigation hint for related diagnostics

Ferdi265:
This PR adds an additional setting to coc.nvim that makes it possible to disable the command hint that is displayed for related diagnostics, since experienced users might like a cleaner diagnostic popup.


fannheyward:
I've removed the hint message, and I think there is no need to add an option for this.

PR #1769, opened 2020-04-16
fix(workspace): don't change documentChanges order

### LINKED ISSUE #1767: Is the ordering of documentChanges respected?
I had someone report some odd behavior in our [coc-metals](https://github.com/scalameta/coc-metals/issues/164) plugin, and I've seemed to have tracked it down to something not working correctly with the way that `DocumentChanges` are processed when inside of a `textDocument/rename`. In Metals, you are able to rename a symbol, and if that symbol for example is a Class, and that filename matches that Class name, the rename will both rename the symbol and also the file name. The response from the server will look like the following if I attempt to do a rename on a Class `Test` in `Test.scala`.

`textDocument/rename`
```json
{
  "documentChanges": [
    {
      "textDocument": {
        "version": null,
        "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/Test.scala"
      },
      "edits": [
        {
          "range": {
            "start": {
              "line": 2,
              "character": 7
            },
            "end": {
              "line": 2,
              "character": 11
            }
          },
          "newText": "Hi"
        }
      ]
    },
    {
      "oldUri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/Test.scala",
      "newUri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/Hi.scala",
      "kind": "rename"
    }
  ]
} 
```

As you can see, there are two `documentChanges`. The first will be the edit to rename the actual symbol, and then second will be to rename the file. In VS Code, this works as expected. However, in both Nvim and Vim this will actually do the rename file first, and then attempt to the rename symbol second. This results in the `Hi.scala`file still having the old `Test` Class name and then a new buffer called `Test.scala` having nothing but the new text in it, `Hi`. This makes me assume that the order of the `documentChanges` aren't happening in the order that they are in the array.

## Result from CocInfo
I've tried this on the following:

```
vim version: NVIM v0.5.0-3d1531a
node version: v12.13.1
coc.nvim version: 0.0.78
term: iTerm.app
platform: darwin
```
And also:
```
vim version: VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Apr 13 2020 07:47:09)
node version: v12.13.1
coc.nvim version: 0.0.78
term: iTerm.app
platform: darwin
```

## Reproduce the bug
This is sort of hard to mimic probably, but you just need a rename that returns an array of `documentChanges`, and it needs to be asserted that the events in that array are processed in order.

## Screenshots (optional)

I've captured this in a gif to demonstrate the behavior I'm talking about.

As always, thanks for your hard work and great tool!

fannheyward:
`TextDocumentEdit` changes should be in first order. @ckipp01 can you test this PR with #1767 ?

close #1767

### DIFF
--- a/src/workspace.ts
+++ b/src/workspace.ts
@@ -1661,7 +1661,7 @@ augroup end`
         res.push(change)
       }
     }
-    res.push(...documentEdits)
+    res.unshift(...documentEdits)
     return res
   }
 


chemzqm:
Doesn't look right.

fannheyward:
`mergeDocumentChanges` will move `TextDocumentEdit` changes to the end of the list. 

```json
{
  "documentChanges": [
    {
      "textDocument": {
        "version": null,
        "uri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/Test.scala"
      },
      "edits": [
        {
          "range": {
            "start": {
              "line": 2,
              "character": 7
            },
            "end": {
              "line": 2,
              "character": 11
            }
          },
          "newText": "Hi"
        }
      ]
    },
    {
      "oldUri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/Test.scala",
      "newUri": "file:///Users/ckipp/Documents/scala-workspace/test-project/example/src/Hi.scala",
      "kind": "rename"
    }
  ]
}
```

from #1767 , after `mergeDocumentChanges`, file rename will become first change, and textEdit as second.

chemzqm:
But `rename` could exists before document edit, I will made the fix.

ckipp01:
> `mergeDocumentChanges` will move `TextDocumentEdit` changes to the end of the list.
> ...
> from #1767 , after `mergeDocumentChanges`, file rename will become first change, and textEdit as second.

I'm not sure that would work. Since if you shift them around, that's exactly what is happening now. The rename happens, but then the `textEdit` uri would need to change, and you probably don't want to start trying to manually edit the uri's in the `textEdits`.

PR #1771, opened 2020-04-16
feat(configuration): add replacement patterns for URIs

pderichai:
This is a bit of a niche feature, but it's something that I absolutely need in order to continue using coc.nvim, and I imagine it could be useful to other people facing similar restrictions.

# Motivation
I recently encountered a problem where the language server I wanted to use expected files to have a certain absolute path to work correctly. For example, imagine that the language server explicitly requires code to live under `/code`. Usually, this isn't a problem because if the language server expects such a path, there's probably a good reason and I can make sure to put code under that path.

Unfortunately, I'm operating under the restriction that the mount point of the filesystem I work with cannot be changed e.g. the filesystem must be mounted under `/Volumes`. I'm also unable to bind another mount point to the filesystem, and vim's `getcwd()` does not respect symlinks.

# What I've Added

I've added a variable `g:coc_uri_prefix_replace_patterns` which is a map of strings where the first pattern will be replaced by the second when getting the URI of a Document.


chemzqm:
Checkout `:h coc_cygqwin_path_prefixes`

pderichai:
Thanks for pointing me to that. I'm not using cygqin, and I'm not using Windows, so using something like `let g:coc_cygqwin_path_prefixes = {'/Volumes': ''}` doesn't work for me. I'm guessing the replacement only happens if I'm also on Windows and using Cygqwin? Do you have any other suggestions?

src/model/document.ts
```diff
@@ -12,6 +12,7 @@ import { byteIndex, byteLength, byteSlice, characterIndex } from '../util/string
 import { Chars } from './chars'
 import { group, distinct } from '../util/array'
 import { comparePosition } from '../util/position'
+import workspace from '../workspace'
```
chemzqm, line 15: Please don't import workspace here

data/schema.json
```diff
@@ -1010,6 +1010,14 @@
         "type": "string"
       }
     },
+    "coc.util.index.uriReplacePatterns": {
```
chemzqm, line 1013: Use `coc.preferences. uriReplacePatterns` instead.

pderichai: Done.

pderichai:
Please reopen this pull request so the changes I've made from your review can be seen.

data/schema.json
pderichai: Done -- you should see this change when the pull request is reopened.

src/model/document.ts
pderichai: I'm not sure how else to get the configuration without workspace. What would you suggest?

chemzqm:
Screen Shot 2020-04-17 at 3 38 10 PM

pderichai:
Sorry, I did a rebase. Can you reopen it now? If not, I'll just make a new PR.

src/util/index.ts
```diff
@@ -25,9 +25,13 @@ export function wait(ms: number): Promise<any> {
   })
 }
 
-export function getUri(fullpath: string, id: number, buftype: string, isCygwin: boolean): string {
+export function getUri(fullpath: string, id: number, buftype: string, isCygwin: boolean,
+                       uriReplacePatterns: [string, string][]): string {
```
fannheyward, line 29: How about get configuration here? Without change model/document.

pderichai: That's fine with me. Can you suggest a way I could get the config without importing workspace?

fannheyward: Import and use workspace to get configuration.

pderichai: Done.

pderichai:
Thanks for the reviews -- I've implemented the suggestions. The current failing tests don't look related to my changes. Let me know if you'd like me to merge in changes to my fork.

fannheyward:
@chemzqm can't use workspace in `util`?

pderichai:
@chemzqm Just bumping to see if there's any more work that needs to be done on this PR.

pderichai:
@chemzqm @fannheyward importing workspace in either util/index.ts and model/document.ts created a circular dependency. I've refactored the feature to be configured through a vim variable, much like how g:coc_cygqwin_path_prefixes works.

fannheyward:
LGTM, cc @chemzqm

PR #1796, opened 2020-04-20
fix(workspace): no loadfile after create file

fannheyward:
I don't think it's need to load file after create file.

1. `applyEdit` with `createFile` change
2. do `createFile` works, it's OK
3. `loadFile` after `createFile`, this will change current `workspace.bufnr` to new file bufnr, it's not correct
4. do `jumpTo` failed, because workspace.bufnr changed

### DIFF
--- a/src/workspace.ts
+++ b/src/workspace.ts
@@ -939,7 +939,6 @@ export class Workspace implements IWorkspace {
         if (doc) return
         let encoding = await this.getFileEncoding()
         fs.writeFileSync(filepath, '', encoding || '')
-        await this.loadFile(uri)
       }
     }
   }


chemzqm:
> loadFile after createFile, this will change current workspace.bufnr to new file bufnr, it's not correct

It should not, I can't reproduce.

fannheyward:
```js
  it('test', async () => {
    const doc = await helper.createDocument()
    expect(workspace.bufnr).toEqual(doc.bufnr)

    const file = path.join(__dirname, uuid())
    const uri = URI.file(file).toString()
    const workspaceEdit: WorkspaceEdit = {
      documentChanges: [CreateFile.create(uri, { overwrite: true })]
    }
    const res = await workspace.applyEdit(workspaceEdit)
    expect(res).toBe(true)
    expect(workspace.bufnr).toEqual(doc.bufnr)
  })
```

chemzqm:
I can reproduce it, it's problem of buffer events.

chemzqm:
Fixed.

PR #1813, opened 2020-04-26
fix(extensions): doc action is missing

fannheyward:
`:CocList extensions` - `Tab` will list extension actions, but `doc` action is missing, and actions after `doc` will return wrong choose number.


fannheyward:
@chemzqm `doc` has been used `d` for disable, `o` for open and `c` for configuration, so `doc` action will be ignored.

PR #1819, opened 2020-04-27
fix(float): don't leak floating windows when created concurrently

### LINKED ISSUE #1816: Empty popup leaked through race in floatFactory
In the regular course of using clangd, coc will eventually "black out" a small 3x1 region in the center of the window:

square

The region is an empty popup created via `floatFactory`, and will **never be cleared** by coc, resulting in "dead cells" in the window.
Because it is never cleared once it occurs, this is a difficult bug to live with despite being rare/flaky/hard to reproduce :-(

I've finally managed to track this down. I don't have a tiny reliable repro but the following help:
 - use vim8, not neovim
 - `autocmd User * sleep 1` to increase the chance of hitting the race
 - perform an edit that generates different diagnostics in sequence as you type

The problem occurs when `FloatFactory.checkFloatBuffer()` is entered twice concurrently. While `await createPopup()` is blocked, we try to process another set of diagnostics, end up back in `checkFloatBuffer()` and call `createPopup()` again. The result of each call is written to `this.popup` so one of the popup window IDs is leaked.
(Confirmed this sequence with logging)

I'm not sure exactly what the best fix is, I'll try to come up with a patch, you might have better ideas.

sam-mccall:
Fixes #1816

### DIFF
--- a/src/__tests__/modules/decorator.test.ts
+++ b/src/__tests__/modules/decorator.test.ts
@@ -1,0 +1,27 @@
+import * as decorator from '../../util/decorator'
+
+class CallTest {
+  public count = 0
+
+  @decorator.combineConcurrent
+  public async combinedConcurrent() : Promise<number> { return ++this.count }
+}
+
+describe('combineConcurrent', () => {
+  test('overlapping', async () => {
+    const c = new CallTest()
+
+    const first = c.combinedConcurrent()
+    const second = c.combinedConcurrent()
+    expect(await first).toBe(1)
+    expect(await second).toBe(1)
+  })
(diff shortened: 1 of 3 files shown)


sam-mccall:
I didn't add a test for the `@memorize` decorator while here because it seems to be broken, so I'll send another PR with fix + test.

chemzqm:
Thanks for contribution, but I tend make some rework with floatFactory module.

chemzqm:
Should be fixed now.

sam-mccall:
Thanks! Looks good so far - my artificial cases can't reproduce it anymore.

PR #1829, opened 2020-04-29
Add virtualtext highlight links according to documentation

TimoFreiberg:
I noticed that the settings described in the documentation aren't actually there, so I added them.
https://github.com/neoclide/coc.nvim/blob/b47c4a8738bcb11c693534a523185baf2c0f9a65/doc/coc.txt#L1885

### DIFF
--- a/plugin/coc.vim
+++ b/plugin/coc.vim
@@ -268,6 +268,12 @@ endif
 if has('nvim-0.5.0')
   hi default CocCursorTransparent gui=strikethrough blend=100
 endif
+if has('nvim')
+  hi default link CocErrorVirtualText   CocErrorSign
+  hi default link CocWarningVirtualText CocWarningSign
+  hi default link CocInfoVirtualText    CocInfoSign
+  hi default link CocHintVirtualText    CocHintSign
+endif
 
 hi default link CocHoverRange     Search
 hi default link CocCursorRange    Search


fannheyward:
We don't need to define these HL link, coc will get HL name by https://github.com/neoclide/coc.nvim/blob/master/src/diagnostic/buffer.ts#L182, `CocErrorVirtualText` is only used for user custom settings.

oblitum:
@fannheyward I'm a bit confused.

With this pull, I indeed get different highlighting, meaning it does cause a change compared to current behavior. But the pull is just providing the default highlighting for those groups according to what the docs states.

If `CocErrorVirtualText` is only used for user custom settings, why say it has a default value that seems of no use?

chemzqm:
It's already defined https://github.com/neoclide/coc.nvim/blob/master/src/diagnostic/manager.ts#L159

oblitum:
For some very unknown reason I don't have [any of the `'Coc'.name.'VirtualText'`](https://github.com/neoclide/coc.nvim/blob/f3e3f9401acd5822adb96e6f435b8be973e9cd30/autoload/coc/util.vim#L989) set.

Trying to figure out why.

oblitum:
Ah ok, figured it out.

The highlighting groups are only getting their default values if `diagnostic.virtualText` is set to true: https://github.com/neoclide/coc.nvim/blob/master/src/diagnostic/manager.ts#L158. But for chaining hints as per https://github.com/fannheyward/coc-rust-analyzer/issues/177, there's no diagnostics involved.

I suggest to remove the check, it's harmless to have the groups defined, even if `diagnostic.virtualText` is `false`.

oblitum:
> I suggest to remove the check, it's harmless to have the groups defined, even if `diagnostic.virtualText` is `false`.

@fannheyward or in case that ^ doesn't happen, maybe disassociate coc-rust-analyzer groups from coc.nvim diagnostic groups.

fannheyward:
@oblitum maybe link `CocRustChainingHint` to `CocHintSign` directly is a better way?

oblitum:
@fannheyward yes it seems, given `CocHintVirtualText` depends on coc virtualText diagnostics feature for it to exist.

oblitum:
@fannheyward though there's a trade-off, the appearance may not be the expected one for those that have coc virtual text diagnostics enable and `CocHintVirtualText` set. Extra logic can be applied to check whether `CocHintVirtualText` exists or not, before setting to `CocHintSign`, though it turns what would be a simple setting into a complex one.

PR #1854, opened 2020-05-05
chore(package): build with local installed webpack

UncleBill:


chemzqm:
Webpack is only used for build release bundle, not for end user, it's not included in package.json to avoid potension issue with webpack.

PR #1855, opened 2020-05-05
Release minified code

UncleBill:
and reduce build size

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ publish.sh
 !src/__tests__/tags
 src/__tests__/extensions/db.json
 package-lock.json
+.DS_Store
--- a/package.json
+++ b/package.json
@@ -69,7 +69,9 @@
     "ts-jest": "^24.2.0",
     "tslint": "^5.19.0",
     "typescript": "^3.8.2",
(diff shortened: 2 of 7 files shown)


chemzqm:
It's intended to not minify code.

UncleBill:
>It's intended to not minify code.

Why not? With minified code, node can boot faster.

oblitum:
Can, but did you measure any relevant effect? Otherwise it's moot.

UncleBill:
>Can, but did you measure any relevant effect?

Haven't.

But according to _Cost of JavaScript [1]_, less code means less time to parse and compile.

[1]: https://v8.dev/blog/cost-of-javascript-2019

oblitum:
It is still moot. This may be solving a non-issue that will make it harder to browse and debug code for no real benefit.

UncleBill:
You can debug with Chrome devtools which also supports source-maps for node debugging.

It's OK if you don't want to ship with minified code, I can drop that commit.

chemzqm:
It might be faster, but much harder for bug trace.

chemzqm:
webpack is only needed for build the bundle script, so It's excluded

PR #1867, opened 2020-05-08
fix(languages): incorrect startcol check

fannheyward:
@chemzqm  typo?  `if (startcol < option.col)` will never work.


chemzqm:
I've made a fix.

PR #1870, opened 2020-05-08
Create pluggable mappings for classes: <Plug>(coc-classobj-*)

pappasam:
Creates pluggable mappings for class text objects. Also, generalizes the function code to make it generally applicable to any DocumentSymbol range.

Topic for discussion: should we consider `['Interface', 'Struct', 'Class']` as all part of a "class" text object? I think about them similarly across languages, so I'm inclined to say yes.

I'll update the English docs if you generally approve of this PR / once we agree on the implementation.

### DIFF
--- a/plugin/coc.vim
+++ b/plugin/coc.vim
@@ -1,8 +1,8 @@
 if exists('g:did_coc_loaded') || v:version < 800
   finish
 endif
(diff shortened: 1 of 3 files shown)


chemzqm:
LGTM

PR #1875, opened 2020-05-09
Document coc-classobj-* textobj

pappasam:
Documents feature provided by https://github.com/neoclide/coc.nvim/pull/1870

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,13 @@ xmap af <Plug>(coc-funcobj-a)
 omap if <Plug>(coc-funcobj-i)
 omap af <Plug>(coc-funcobj-a)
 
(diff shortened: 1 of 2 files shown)


oblitum:
@pappasam there's no default mapping as far as I can see (which is good), so the docs are not correct stating `ac` and `ic` are default mappings?

oblitum:
They're solely mentioned in README _example_ configuration.

pappasam:
You’re right! My apologies, I was blindly following the existing examples to document the class objects

pappasam:
Guess both the old and new docs need to be updated.

oblitum:
> My apologies

Not a problem :)

PR #1896, opened 2020-05-12
fix(extensions): fix coc_global_extensions error

### LINKED ISSUE #1881: Latest release commit (8909a94) broke coc_global_extensions
They just don't work on a fresh install. No extensions get installed. Nothing in the package.json.

When I revert to `244b439` they work just fine.

Sorry for not filling out all the details as typical in a bug report, but this should be easily reproducible.

fannheyward:
close #1881

@chemzqm correct me if I'm wrong, what `checkExtensions` does is:

1. get globalExtensions and filter disabled
2. check if exists in node_modules
3. check json.dependencies, here I think it's only need to check whether exists
4. finally got non-exist extensions list, do install action

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -180,7 +180,7 @@ export class Extensions {
       let json = this.loadJson()
       if (json && json.dependencies) {
         let vals: string[] = Object.values(json.dependencies)
-        names = names.filter(s => /^https?:/.test(s) && vals.some(v => v.startsWith(s)))
+        names = names.filter(s => !vals.includes(s))
       }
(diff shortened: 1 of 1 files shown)


chemzqm:
Url could have tags, so `includes` can't be used.

fannheyward:
CocInstall doesn't support url with tags installation.

chemzqm:
It's added by npm like `#master`

fannheyward:
`names = names.filter(s => !vals.some(v => v.includes(s)))` ?

Issue #1008, opened 2019-07-21
[Feature request] Add a way to reverse the order of results with CocList


malob:
**Is your feature request related to a problem? Please describe.**
When searching through lists with `coc.nvim`, I find it a little annoying that the best matchs are placed at the top since it's furthest from where I'm typing (and therefore looking most of the time). Auto-resizing (which I like) adds to the annoyance since the location of the best match moves on the screen as I type. 

**Describe the solution you'd like**
I'd love `:CocList` to have an option like `--reversed` that would reverse the order of the list so that the best matches appear at the bottom. Ideally, there'd also be an easy way to make this the default for all lists (maybe by adding an option to the `coc.nvim` configuration like `list.reversed` which would default to `false`).

**Additional context**
This request is inspired by the same functionality in [`denite.nvim`](https://github.com/Shougo/denite.nvim/blob/3a1b508116652a6c9b4b174ddafaee7852a15fc4/doc/denite.txt#L818).

chemzqm:
PR welcome.

vantreeseba:
I have this started, but I have been busy.

I do plan on finishing and opening PR at some point.

https://github.com/vantreeseba/coc.nvim/tree/list_reverse_items

nomasprime:
Also missing this functionality from denite.

snprajwal:
Will PR for this still be accepted? @chemzqm

PR #1905, opened 2020-05-13
Reversed autocomplete options list

hrai:
Added `suggest.reverseList` setting to reverse the order of autocomplete options so most relevant ones show at the bottom instead of top.

Resolves https://github.com/neoclide/coc.nvim/issues/1008

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -425,6 +425,11 @@
       "description": "Remove completion items with duplicated word for all sources, snippet items are excluded.",
       "default": false
     },
+    "suggest.reverseList": {
+      "type": "boolean",
+      "description": "Reverse completion items.",
+      "default": false
+    },
     "suggest.defaultSortMethod": {
       "type": "string",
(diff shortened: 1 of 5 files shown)


chemzqm:
How could it be useful?

hrai:
@malob has described the use-case here. I had similar needs and came across this issue.
https://github.com/neoclide/coc.nvim/issues/1008#issue-470736663

I used to use YCM. It allowed bottom up sorted list so the most relevant items were closest to the cursor and not at the top.

chemzqm:
It's not list of complete items.

hrai:
It's based on the same scenario except that this PR is for the list of complete options. 

In the screenshot below, it would be helpful if the most relevant item was closest to the cursor (which is at the bottom of the popup) rather than at the top.
image

chemzqm:
@hrai it's not helpful, you won't see it when there is scrollbar.

hrai:
Alrighty, I'll close it.

PR #1910, opened 2020-05-13
Add subseparator between codelenses in virtual text

hhiraba:
Added sub-separator between codelenses in virtual text in case there are multiple codelenses at one line.

The motivation is that [`rust-analyzer`](https://github.com/rust-analyzer/rust-analyzer) returns CodeLens item whose title includes space character like `Run Bench`. If it appears with other CodeLenses, virtual text becomes confusing.

Screenshot from 2020-05-14 00-47-50

So I added `codeLens.subseparator` option. 
If `codeLens.subseparator: " | "`, virtual text in same case as above is displayed as below:

Screenshot from 2020-05-14 00-48-24

(If `codeLens.subseparator: " "`, behaves as before)

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -668,6 +668,11 @@
       "description": "Separator text for codeLens in virtual text",
       "default": "‣"
     },
+    "codeLens.subseparator": {
+      "type": "string",
+      "description": "Subseparator between codeLenses in virtual text",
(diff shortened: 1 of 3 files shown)


hhiraba:
Sorry I did not update tests. I will do later.

PR #1946, opened 2020-05-18
add suggest.maxPreviewHeight

tomtomjhj:
This is a configuration to limit the max height of floating windows AND
preview windows. This is especially useful when hoverTarget is preview
and previewAutoClose is false.  The Main advantage of such configuration
is that the documentation text remains intact while editing/navigating.
However, if the documentation is very long, the preview window would
occupy the entire page:
```
// in provideTextDocumentContent
nvim.command(`exe "normal! z${this.documentLines.length}\\<cr>"`, true)
```
Users can prevent such cases by setting maxPreviewHeight to a value
small enough (e.g. 20).

The name of option doesn't look good because there is another option
with the exactly same name (list.maxPreviewHeight).  Something like
suggest.hoverMaxHeight and suggest.hoverMaxWidth would be better.

**TODO**:
It turns out that this actually doesn't limit the max height of floating
window and I couldn't figure out why.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -277,10 +277,15 @@
       "description": "Enable preselect feature, only works on neovim",
       "default": false
     },
+    "suggest.maxPreviewHeight": {
+      "type": "number",
+      "default": 40,
+      "description": "Maximum height of floating window or preview window."
+    },
     "suggest.maxPreviewWidth": {
       "type": "number",
       "default": 80,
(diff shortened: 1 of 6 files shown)


fannheyward:
cc @chemzqm

chemzqm:
It is confusing, you call it `suggest. maxPreviewHeight` but use it for float window of hover and float window of popup menu.

Issue #2019, opened 2020-06-02
Disable certain sources on filetype or buffer level


kristijanhusak:
**Is your feature request related to a problem? Please describe.**
I'm not sure if this is already possible. I know i can do `b:coc_enabled = 0`, but i don't want to disable whole thing. 
For example, I'm using `coc-tag` by default almost everywhere. Only place where i don't need it is when writing sql. For that filetype I would like to disable it.

**Describe the solution you'd like**
I would like to have an option either to blacklist or whitelist sources on buffer level. For example, in my case, something like this:
```vimL
" blacklist
let b:coc_disabled_sources = ['coc-tag']

" or whitelist
let b:coc_enabled_sources = ['coc-db']
```

**Describe alternatives you've considered**
I guess alternative would be to add a whitelist of all filetypes that I use for the coc-tag configuration (`coc.source.tag.filetypes`), but that list can be pretty big.

**Additional context**
I can try to do this myself if you agree with something like this, or you want some other solution that would make this work.

chemzqm:
If you use tag completion a lot, use vim's builtin completion for tag would be better option, we only have whilelist for now

chemzqm:
No such support, it should not be hard to support b:coc_disabled_sources

kristijanhusak:
Yeah i think having coc_disabled_sources is better than coc_enabled_sources. Whitelist can be easily achieved with the current functionality.

Should I leave it to you or should I try it?

chemzqm:
I prefer you try it out

chemzqm:
I've added support for `g:coc_sources_disable_map` fd6f1a5e.

z0xyz:
@chemzqm Is this variable still viable? It doesn't seem to be anywhere at the docs, whether that the `coc-variables` or `cco-nvim.txt`. It also seems to be ineffective whether set or not

chemzqm:
It's removed, use b:coc_disabled_sources instead

z0xyz:
It's rather quirky for me i.e. let's say for any buffer of filetype `text`, i set the variable set in this way `let b:coc_disabled_sources = ['around', 'buffer' , 'file']`.  When i open a text file, and input a `/` forward slash character, it still shows path completions, but it only allows tabbing, but no substring matching of the completions available . Same behavior applies for every other disabled source. They show up when their trigger character is inputted, but they only allow tabbing .

z0xyz:
@chemzqm Should i preferably open an issue instead ?

[...]

PR #2020, opened 2020-06-02
Add disabled sources option.

kristijanhusak:
This is a fix  for #2019.


kristijanhusak:
I have no idea why tests fails, but they are failing for me also on master branch, so it looks like it's not related to my changes.

chemzqm:
Sources module need some rework, fixed by fd6f1a5e

kristijanhusak:
Thanks!

PR #2044, opened 2020-06-08
Fix drawing horizontal line on CJK

tasuten:
This reverts commit 915af62ce70cef1dffe04dc63a44763d9706fec7.

On environments that East-Asian `Ambiguous characters are double-width`[1],
`\u2500` is treated as a double width character, and 
horizontal line in popup will be drawn bad like below screenshot.
(Screenshot on iTerm2, macOS)

double-width

`-` is treated as a single width character anytime, so it will be drawn good.
pull-req

[1] UAX # 11: East Asian Width http://www.unicode.org/reports/tr11/tr11-38.html


tasuten:
Oh sorry, test code still uses `\u2500`, so test failed.

fannheyward:
@tasuten test has been fixed.

PR #2083, opened 2020-06-17
fix(workspace): Auto jump to first location

sheerun:
If there is only one result available then it's auto-jumping to location, but not when there are multiple ones. This fixes it.

### DIFF
--- a/src/workspace.ts
+++ b/src/workspace.ts
@@ -705,6 +705,7 @@ export class Workspace implements IWorkspace {
     if (preferences.get<boolean>('useQuickfixForLocations', false)) {
       await nvim.call('setqflist', [items])
       nvim.command('copen', true)
+      nvim.command('cfirst', true)
     } else {
       await nvim.setVar('coc_jump_locations', items)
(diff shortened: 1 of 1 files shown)


chemzqm:
It's intended, you have to use autocmd if you would want jump.

sheerun:
Is this documented? I don't know how

chemzqm:
`:h QuickFixCmdPost`

sheerun:
QuickfixCmdPost doesn't work because it's not quickfix command that opens quicklist, but copen

chemzqm:
Use `BufWinEnter` autocmd.

sheerun:
While following echoes ok when opening quicklist:

```
autocmd BufWinEnter quickfix echo "ok"
```

Following doesn't do anything:

```
autocmd BufWinEnter quickfix cfirst
```

sheerun:
@chemzqm https://stackoverflow.com/questions/62467941/how-to-configure-vim-to-automatically-open-first-quicklist-item?noredirect=1#comment110480377_62467941

PR #2116, opened 2020-06-27
fix: when insertText is empty

weirongxu:
Coc should remove the candidate when insertText is empty.

https://github.com/fannheyward/coc-docthis/blob/ecadbce31ec5ea673e5627df7ef07ac2178ad4ce/src/index.ts#L41

### DIFF
--- a/src/languages.ts
+++ b/src/languages.ts
@@ -617,7 +617,7 @@ class Languages {
         let item = completeItems[vimItem.index]
         if (!item) return
         let line = opt.linenr - 1
-        if (item.insertText && !item.textEdit) {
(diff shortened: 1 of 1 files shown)


chemzqm:
Could you explain?

weirongxu:
Sorry, does the type of insertText include the `null`? Because I see that the type is `string?`,  I thought it is undefined if insertText was not provided.

chemzqm:
No, but some language server would return null.

weirongxu:
I change this because when my insertText is empty, but the label is not empty, coc.nvim should replace the inserted label with empty then execute the other command.

image

PR #2134, opened 2020-07-02
fix(workspace): fire QuickfixCmdPost for quickfix

sheerun:
This is alternative solution for #2083


chemzqm:
I don't think we should trigger QuickfixCmdPost, checkout c9246e45

sheerun:
Thanks!

PR #2139, opened 2020-07-03
FIX(completion): Add language ids to client name 

### LINKED ISSUE #2138: Does not support multiple completion sources from a single LSP source.
https://github.com/neoclide/coc.nvim/blob/master/src/language-client/client.ts#L1885

Due to using the client id here, when more than one type of file completion is registered from a server, it overrides the previous one,

Since this is just a lookup string, can we make this unique by doing something like 

this._client.id + languageIds so that if a server provides capabilities for completion for different language ids, it correctly registers?

This is the cause of vshaxe/vshaxe#447

vantreeseba:
to prevent sources with the same name being created when a server registers more than one completion source.

closes #2138
closes vshaxe/vshaxe#447
closes vantreeseba/coc-haxe#5


chemzqm:
Fixed by aa8ffe0e

PR #2141, opened 2020-07-04
Bugfix strengthen is document edit type guard

fnune:
### Disclaimer

- I'm not aware of the performance impacts this could have. This is just an exploratory PR to fix a bug.
- Some tests fail locally with or without my changes, so I refrained from editing the test suite. I need some guidance on this.

### Context

I was trying to run my binding on a Svelte file:

`command! -nargs=0 Prettier :call CocAction('runCommand', 'prettier.formatFile')`

And I was getting this `Cannot find name 'replace' of undefined` error. After setting up the runtime path to my local coc.nvim installation, I found out it's because of a naïve check from `isDocumentEdit`. The `TextEdit`s inside it are asserted by `as TextDocumentEdit` as all having a `.newText` string field, but some were `undefined`.

This pull request does three things:

- Strenghtens `isDocumentEdit` checks by checking nested fields
- Removes unnecessary `as TextDocumentEdit` assertions
- Fix applying edits with `TextEdit`s that were missing `.newText`

After these commits, the message from coc.nvim became `[coc.nvim] svelte not supported by prettier`. It's better than `Cannot find name 'replace' of undefined`.

Thanks for the awesome software, it's changed my vim productivity for the better!

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -8,7 +8,7 @@ import { BufferOption, ChangeInfo, DidChangeTextDocumentParams, Env } from '../t
 import { distinct, group } from '../util/array'
 import { diffLines, getChange } from '../util/diff'
 import { isGitIgnored } from '../util/fs'
-import { disposeAll, getUri } from '../util/index'
+import { disposeAll, getUri, isTextEdit } from '../util/index'
 import { comparePosition } from '../util/position'
 import { byteIndex, byteLength, byteSlice } from '../util/string'
 import { Chars } from './chars'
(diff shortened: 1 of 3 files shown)


fnune:
It's strange, the test that's failing on CI is also failing locally for me on `master`.

fnune:
My initial problem persists. My `Prettier` command shows `[coc.nvim] svelte not supported by prettier`. However, formatting on save works perfectly, so this is surprising.

```
  "coc.preferences.formatOnSaveFiletypes": ["svelte"],
```

chemzqm:
Fixed on coc-prettier

fnune:
Awesome! Thanks

PR #2161, opened 2020-07-08
fix(diagnostic): Allow messages in INSERT mode

tombh:
Relates to #810


chemzqm:
No, it would cause unexpected behavior.

tombh:
So you won't support displaying diagnostic messages in Insert mode?

chemzqm:
Not automaticlly. You can check `truncate` argument to make sure it's invoked by user.

tombh:
This line? `if (truncate && workspace.insertMode) return` I don't understand what `truncate` is for. I don't understand most of this code, so maybe I'm taking the wrong approach?

I just want to be able to see diagnostic messages in Insert mode :'(

PR #2220, opened 2020-08-01
fix: resolveRoot in windows

weirongxu:
`os.homedir()` get the uppercase disk symbol `C:`, but dir is lowercase


weirongxu:
Open a full path file, `nvim C:/Users/Deafult/repos/demo`
wile resolve a home directory, because the home directory contains `.vim`, for example on this rootPatterns https://github.com/coc-extensions/coc-omnisharp/blob/master/package.json#L94

PR #2229, opened 2020-08-04
perf(autocmd): remove calls to get() for 15% faster autcmd

antoinemadec:
It shows on CursorMoved autocmd especially, here are profiling results:
- 1295   0.918150   0.122849  <SNR>40_Autocmd() (before)
- 1289   0.821687   0.092443  <SNR>40_Autocmd() (after)

It aint much, but it's honest work.

### DIFF
--- a/plugin/coc.vim
+++ b/plugin/coc.vim
@@ -47,6 +47,7 @@ endfunction
 call s:checkVersion()
 
 let g:did_coc_loaded = 1
+let g:coc_workspace_initialized = 0
 let g:coc_service_initialized = 0
 let s:is_win = has('win32') || has('win64')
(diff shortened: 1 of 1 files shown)


oblitum:
@antoinemadec I think there's a reason it is like that, it's because they may be called before the plugin is fully initialized.

antoinemadec:
The function is defined after the variable. How can this function be called if it's not defined? If defined, the global variable will be too (i set it to 0 at the beginning of the file).

oblitum:
@antoinemadec I've seen some vimscript behavior, specially when calling a function from command line (nvim +MyCommand), where you can call a function where the script itself where it's defined did not run. It may or may not be due to that.

PR #2254, opened 2020-08-17
fix: coc-diagnostic-next follows diagnostic.level

fannheyward:
cc @chemzqm


chemzqm:
> `if (level) diagnostics = diagnostics.filter(o => o.severity <= level)`

Would break `coc-diagnostic-next-error` `coc-diagnostic-prev-error`

PR #2264, opened 2020-08-19
Add jumpTimeout configuration

### LINKED ISSUE #2260: Support timeout configuration for jumpDefinition and jumpImplementation

I am working on Kubernetes and using gopls as my LSP via [coc-go](https://github.com/josa42/coc-go). Kubernetes is a monorepo with loads of dependencies so gopls seems to take some time to spin up and process requests.

I often get the error `[coc.nvim]: Error on notification "jumpDefinition": getDefinition timeout after 5s` when trying to jump to a definition or pull up the implementation before the file is cached.

I'd like to be able to configure the timeout for these requests like I can configure `suggest.timeout`.

The alternative right now is to wait a bit after the timeout for the LSP to eventually load the file.

**Additional context**

I believe this is the hardcoded timeout that needs to be made configurable.

https://github.com/neoclide/coc.nvim/blob/e9ce9768e257432d021a90b5cbe00fb4b35e30a4/src/languages.ts#L70I

eddiezane:
Closes #2260.

I took a shot at adding this configuration. It probably shouldn't live under `suggest` and `jumpTimeout` might not be clear so I can rename it and move it where ever we decide.

It will also need a proper translation for the docs once decided on.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -392,6 +392,13 @@
       "maximum": 15000,
       "description": "Timeout for completion, in miliseconds."
     },
+    "suggest.jumpTimeout": {
+      "type": "integer",
+      "default": 5000,
+      "minimum": 500,
+      "maximum": 15000,
+      "description": "Timeout for jumps, in miliseconds."
+    },
     "suggest.minTriggerInputLength": {
       "type": "number",
       "default": 1,
--- a/doc/coc.cnx
+++ b/doc/coc.cnx
@@ -116,6 +116,10 @@ NodeJS https://nodejs.org/ >= 10.12.0
 <
 补全开启时选中第一项： >
 
+	"suggest.jumpTimeout.TODO.TRANSLATION": 500,
+<
+补全开启时选中第一项： >
+
 	"suggest.noselect": false,
 
 保留使用当前的 `completeopt` ：
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -220,6 +220,11 @@ Built-in configurations:~
 
 	Timeout for completion (unit: milliseconds), default: `5000`
 
+"suggest.jumpTimeout":~
+
+	Timeout for jumping to definitions or implementations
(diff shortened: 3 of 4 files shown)


chemzqm:
It doesn't belong to suggest and not only jump affected as well.

four0000four:
As someone who uses CoC as the daily driver for navigating a large codebase, making the hardcoded timeout in [`check()`](https://github.com/neoclide/coc.nvim/blob/e9ce9768e257432d021a90b5cbe00fb4b35e30a4/src/languages.ts#L55) configurable would be a huge time saver.

@chemzqm I agree that this implementation doesn't match the scope of the timeout itself. What if there were a set of configurable timeouts that the user could specify for different `get...` functions, which fall back to a global default if unspecified in the user's config files? Something like:

```typescript
function check(...) {
  let fn = descriptor.value;
  // ... validation of fn ...
  const timeout = getLanguageServerCallTimeout(this, fn);
  // ... use timeout below ...
}

function getLanguageServerCallTimeout(languages: Languages, fn: Function): number {
  const configKey = `suggest.${fn.name}`;
  const { completeConfig } = languages;

  if (completeConfig[configKey]) return completeConfig[configKey];

  const defaultTimeout = 5000;
  logger.debug(`LSP call for ${fn.name} is not configured, defaulting to ${defaultTimeout}`)
  return defaultTimeout;
}
```
This way:
* curious users get to extend timeouts if they desire
* documentation can be added later if necessary
* strict requirements on which functions can be configured can be added later if necessary

If it sounds good to you, and @eddiezane you're busy, I'm happy to try my hand at this.

eddiezane:
@four0000four if @chemzqm green lights your suggestions I'll gladly toss it to you!

I've been chugging along with my local modifications but would be good to get something like this upstream.

chemzqm:
No timeout for location jump any more.

eddiezane:
Thank you for the change!

PR #2283, opened 2020-08-27
fix: diagnostic sign doesn't disappear

### LINKED ISSUE #131: diagnostic sign doesn't disappear
**Describe the bug**
diagnostic sign doesn't disappear

**To Reproduce**

1: test.ts
```javascript
function test() {
  return 0
}
```
2: delete `}`

```javascript
function test() {
  return 0

```
3: insert `}`
```javascript
function test() {
  return 0
}
```

and diagnostic sign doesn't disappear

xuanduc987:
fixes #131

The issue #131 had been fixed before by calling checkSigns on write with commit b300aea2cca22a9c9271ab5569b85b435327b808
But this fix had been reverted after commit c83813f73007d00d5590495d8d6420ad3ded632d

This commit tries to fix the issue in a different way: using sign group.

### DIFF
--- a/src/__tests__/modules/diagnosticBuffer.test.ts
+++ b/src/__tests__/modules/diagnosticBuffer.test.ts
@@ -77,7 +77,7 @@ describe('diagnostic buffer', () => {
     let buf = await createDiagnosticBuffer()
     buf.addSigns(diagnostics)
     await helper.wait(30)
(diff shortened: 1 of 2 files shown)


fannheyward:
This will break CocErrorHighlight.

PR #2295, opened 2020-09-01
feat: list workspace commands only

### LINKED ISSUE #2290: commands for only active extension
I'm always frustrated when i see the commands of inactive extensions...

I would like to have an entry in CocList that only show command's of active extensions... 

couldn't find any alternative solution...

fannheyward:
BREAKING CHANGE: `CocList commands` will list workspace commands only.

close #2290

### DIFF
--- a/src/list/source/commands.ts
+++ b/src/list/source/commands.ts
@@ -32,22 +32,13 @@ export default class CommandsList extends BasicList {
     let list = commandManager.commandList
     let { titles } = commandManager
     let mruList = await this.mru.load()
-    for (let key of titles.keys()) {
+    for (const o of list) {
+      const { id } = o
       items.push({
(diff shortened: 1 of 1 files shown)


fannheyward:
cc @antoinemadec , please check this break coc-fzf or not.

antoinemadec:
@fannheyward , I tried it on a Lua file and Python file.

Both ```CocList commands``` and ```CocFzfList commands``` (using ```CocAction('commands')```) worked as expected.

This is a great feature btw, thanks!

iamcco:
@fannheyward 

It would be better to have an option to change the behavior, because some extension can be active after the command run.

https://github.com/iamcco/coc-flutter/blob/master/package.json#L34
https://github.com/iamcco/coc-flutter/issues/63

PR #2303, opened 2020-09-04
fix: LS env should overwrite process env

fannheyward:
Object.assign later sources will overwrite earlier ones.


ericwq:
@fannheyward can you confirm that this PR fix the TMPDIR env problem of coc.nvim. without this PR, coc.nvim will start go language server with different TMPDIR variable.

fannheyward:
@ericwq yes. When you use `-remote auto`, there should be only one gopls server.

ericwq:
Great! thanks.

PR #2306, opened 2020-09-04
zN<CR> only if preview window is a horizontal split

tomtomjhj:
When using "coc.preferences.hoverTarget": "preview", coc sets the preview window height with `zN<CR>`. The problem is that it reduces the height of the window even if there is no window adjacent to the top/bottom current window, which is undesirable. Therefore, it first need to check that there is a adjacent window.

For example this is what you'll get if you move the preview window to the right size and then do hover.
image


tomtomjhj:
Note: Direction parameter for `winnr` was added in [vim 8.1.1140](https://github.com/vim/vim/commit/46ad288b9b2a6eb0430cf802ff5ce68a58629897), which is higher than minimal version requirement for coc (8.0.1453).

PR #2332, opened 2020-09-12
Add link to my detailed article on coc extensions

pappasam:
I wrote this article because I struggled a bit writing my first coc
extension to wrap an executable language server. The article provides
both the reason and the method to write a coc.nvim extension.
Additionally, it goes over why coc.nvim is different from other nvim LSP
clients and how it relates to vscode. My hope is that it makes it easier
for other developers to write coc extensions so that we get more of
them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,8 +250,8 @@ nnoremap <silent><nowait> <space>p  :<C-u>CocListResume<CR>
 
 - [coc.nvim 插件体系介绍](https://zhuanlan.zhihu.com/p/65524706)
 - [CocList 入坑指南](https://zhuanlan.zhihu.com/p/71846145)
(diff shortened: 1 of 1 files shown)


oblitum:
@pappasam awesome article, great that it introduces beginners to what LSP is, they often look lost what is what on that. Link to it should maybe worth be listed on relevant wiki sessions as well.

oblitum:
@pappasam just some technical criticism though, it may sound at times a bit romantic, from what I can tell, the project emerged from @chemzqm efforts along originally, other great contributor as @fannheyward, only joined quite later, so your article is not correct on that point.

pappasam:
@oblitum I'll update the article with that fact. Thanks for catching!

PR #2412, opened 2020-10-04
feat(util): support getting size for unnamed buffers

yan12125:
With this patch, large unnamed buffers no longer hangs if
`maxFileSize` is correctly configured.

Inspired by https://coderwall.com/p/ussjrg/vim-get-buffer-byte-size


yan12125:
Hmm, I cannot reproduce the two failures from `colors.test.ts` on my machine.

chemzqm:
Need check if the buffer is current one, I've made a fix.

yan12125:
Great, thanks!

PR #2414, opened 2020-10-06
fix: adjust copen height when enable useQuickfixForLocations

kevinhwang91:
default height of `copen` is 10, if the items are less than 10,
we should set height manually to make the behavior like default coclist's.


src/workspace.ts
```diff
@@ -719,7 +719,9 @@ export class Workspace implements IWorkspace {
     const preferences = this.getConfiguration('coc.preferences')
     if (preferences.get<boolean>('useQuickfixForLocations', false)) {
       let openCommand = await nvim.getVar('coc_quickfix_open_command') as string
-      nvim.command(typeof openCommand === 'string' ? openCommand : 'copen', true)
```
kevinhwang91, line 722: I have no idea why execute two times of `openCommand`. Could I delete it?

PR #2465, opened 2020-10-16
fix(float): fix wrong relative position between floatwin and border

voldikss:
Because `anchor` field is not used in window config, the `row` of inner window should be always be larger than that of the outer window

This PR fixes the wrong relative position between those two windows when cursor is at the bottom of the editor screen. 
image
and
image


chemzqm:
Position of border window is wrong, not content window. I've made a fix.

PR #2482, opened 2020-10-20
Fixed incorrect completion suggestions in TabNine

yuki-ycino:
Fixed an issue where there were very few candidates for coc-tabnine completion.
A word in tabnine contains the same thing, so I think it's correct not to check dup if source is tabnine.
Is the existing implementation the intended one?

There are very few completion candidates for coc-tabnine, but with this change, the result is the same as VSCode.

VSCode:
スクリーンショット 2020-10-20 17 43 03

Before:
スクリーンショット 2020-10-20 17 42 51

After:
スクリーンショット 2020-10-20 17 53 17

### DIFF
--- a/src/completion/complete.ts
+++ b/src/completion/complete.ts
@@ -191,7 +191,7 @@ export default class Complete {
         if (asciiCharactersOnly && !/^[\x00-\x7F]*$/.test(word)) {
           continue
         }
-        if ((!item.dup || source == 'tabnine') && words.has(word)) continue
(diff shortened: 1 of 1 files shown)


yuki-ycino:
Related issue: https://github.com/neoclide/coc-tabnine/issues/48

yuki-ycino:
@chemzqm It looks like the dup logic operation is wrong in the PR I got merged earlier. Sorry.
Could you fix `item.dup` to `!item.dup` ?

yuki-ycino:
Thanks!

Issue #2464, opened 2020-10-16
Escaping occuring in hover window


ldelossa:
## versions

vim version: VIM - Vi IMproved 8.2 8021805
node version: v12.18.3
coc.nvim version: 0.0.79-d4d93dc25f
coc.nvim directory: /home/ldelossa/.local/share/nvim/plugged/coc.nvim
term: dumb
platform: linux

## Output channel: go

[Info  - 12:25:42 AM] 2020/10/16 00:25:42 Handshake session update
	update_session=10
	debug_address=""
	logfile=""
	server="1"
	gopls_path="/home/ldelossa/.config/coc/extensions/coc-go-data/bin/gopls"

[Info  - 12:25:42 AM] 2020/10/16 00:25:42 go env for /home/ldelossa/git/redhat/quay-terraform
(root /home/ldelossa/git/redhat/quay-terraform)
(go version go version go1.15.2 linux/amd64
)
(valid build configuration = true)
(build flags: [])
GOMOD=/home/ldelossa/git/redhat/quay-terraform/go.mod
GOMODCACHE=/home/ldelossa/git/gopath/pkg/mod
GOPATH=/home/ldelossa/git/gopath
GOSUMDB=sum.golang.org
GOCACHE=/home/ldelossa/.cache/go-build
GOPRIVATE=
GOPROXY=https://proxy.golang.org,direct
GOFLAGS=
GONOSUMDB=
GOROOT=/usr/local/go
GO111MODULE=
GOINSECURE=
GONOPROXY=

[Info  - 12:25:42 AM] 2020/10/16 00:25:42 go/packages.Load
	snapshot=0
	directory=/home/ldelossa/git/redhat/quay-terraform
	query=[./... builtin]
	packages=2

[Info  - 12:25:50 AM] 2020/10/16 00:25:50 go/packages.Load
	snapshot=4
	directory=/home/ldelossa/git/redhat/quay-terraform
	query=[github.com/quay/quay-terraform]
	packages=1

[Info  - 12:25:50 AM] 2020/10/16 00:25:50 go/packages.Load
	snapshot=4
	package_path="github.com/quay/quay-terraform"
	files=[/home/ldelossa/git/redhat/quay-terraform/config.go /home/ldelossa/git/redhat/quay-terraform/main.go /home/ldelossa/git/redhat/quay-terraform/packer.go]

[Info  - 12:25:57 AM] 2020/10/16 00:25:57 go/packages.Load
	snapshot=7
[...]

PR #2489, opened 2020-10-21
popup: escape markdown syntax

ldelossa:
this commit removes the escapes from markdown returned from a lsp on
hover.

the timing of CocFloatOpen will not work. Instead I tie an autocmd to
the action and call it when the rpc returns.

this pr opens up the use of registering autocmd(s) named after a
CoCAction and having those autocmd fire when the rpc request actually
returns to (n)vim.

### DIFF
--- a/autoload/coc.vim
+++ b/autoload/coc.vim
@@ -11,6 +11,40 @@ let s:warning_sign = get(g:, 'coc_status_warning_sign', has('mac') ? '⚠️ ' :
 let s:select_api = exists('*nvim_select_popupmenu_item')
 let s:callbacks = {}
 
+autocmd User doHover call coc#conceal() <cr>
+
+function! coc#conceal() abort
+  let popup_id = g:coc_last_float_win
+  let bufnr = winbufnr(popup_id)
+  let escapes = []
+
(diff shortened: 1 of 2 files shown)


ldelossa:
closes #2464

chemzqm:
I will not merge this, there could be better solution.

ldelossa:
@chemzqm I'm following your recommendations in issue #2464

chemzqm:
You can use it, but it's hacky and far more from reliable.

ldelossa:
@chemzqm what is better solution? so far I implemented function you said too in issue, and then CocOpenFloat is not usable since it runs before content is in buffer it appears.

chemzqm:
There're many problems like it only works on vim and create autocmd in autoload vimscript, it hack coc#rpc#request to trigger autocmd, but user could use coc#rpc#notify.
I don't have time to explain details, sorry.

chemzqm:
I have said in gitter that we could avoid send unnecessary characters to vim at all by parse higlights and remove those kind of characters.

ldelossa:
Alright, I guess I will just disable mark down for now, as its not usable IMO.

PR #2504, opened 2020-10-23
feat(completion): Allow languageserver level settings for 'suggest'

### LINKED ISSUE #1800: Support language level settings like vscode
Vsocde allows for language level configuration like:
```
{
  "[python]": {
    "editor.codeActionsOnSave": {
      "source.organizeImports": false
    }
  }
}
```
This is not possible in coc-settings.json (well the linter complains at least)

Support vscode like language level configuration

Manually changing it by language

**Additional context**

pedromss:
Close #1800 

- Set listeners for `BufEnter` and `FileType` to update the current `getConfig` function
- Transform the config read from the file from this:

```jsonc
{                                       
   "suggest.autoTrigger": "always",
   "languageserver": {
     "golang": {
       "suggest.autoTrigger": "none", 
        // ...
      }
   }
}
```
into this
```jsonc
{                                       
   "suggest": {
     "autoTrigger": "always"
     "go": {
        "autoTrigger": "none"
     }
   }
   "languageserver": { 
     "golang": {     
       "suggest.autoTrigger": "none", 
       // ...      
      }                
   }               
}
```

@chemzqm need help with:
1. Doesn't take effect when we open with `nvim main.go` for example. Haven't figured out which event gets triggered in this case, do you know?
2. Notice there's a test in `configurations.test.ts`
```
 describe('parse configuration', () => {
    it('should only split top level dot keys', () => { ... }}) 
```

Does something else break if we split all dotted keys? It would be useful to reuse this to break the keys inside:
```jsonc
{                                       
   "suggest.autoTrigger": "always",      
   "languageserver": { 
     "golang": {
       "suggest.autoTrigger": "none", 
        // ...      
      }                
   }               
}
```

I can move the code to specifically split the "suggest" configs inside each `langserver` but its more generic to reuse what already exists, thoughts?

Still need to check if docs need updating and commit the updated tests. But would like feedback on points 1. and 2. above first

### DIFF
--- a/src/completion/index.ts
+++ b/src/completion/index.ts
@@ -44,6 +44,8 @@ export class Completion implements Disposable {
     events.on('InsertEnter', this.onInsertEnter, this, this.disposables)
     events.on('TextChangedP', this.onTextChangedP, this, this.disposables)
     events.on('TextChangedI', this.onTextChangedI, this, this.disposables)
+    events.on('BufEnter', this.onBufEnter, this, this.disposables)
+    events.on('FileType', this.onFileType, this, this.disposables)
     let fn = throttle(this.onPumChange.bind(this), workspace.isVim ? 200 : 100)
     events.on('CompleteDone', async item => {
       this.currItem = null
@@ -65,6 +67,18 @@ export class Completion implements Disposable {
         this.config = this.getCompleteConfig()
       }
     }, null, this.disposables)
+  }
+
+  private async onFileType(ft: string): Promise<void> {
+    this.config = this.getCompleteConfig(ft)
+  }
+
+  private async onBufEnter(bufnr: number): Promise<void> {
+    const doc = workspace.getDocument(bufnr)
+    if (doc == null) {
+      return
+    }
+    this.config = this.getCompleteConfig(doc.filetype)
   }
 
   private get nvim(): Neovim {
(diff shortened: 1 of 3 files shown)


pedromss:
Just noticed the problem with splitting on every `.` is the case where the key is like a url. Is that a real use case tho?

chemzqm:
Should be language specific configuration instead of language server, see #1800

pedromss:
Ok. To clarify, were talking about the configurations that appear in `:h coc-configurations` correct?

For example:
```
{                                       
   "suggest.autoTrigger": "always",
   "languageserver": {
     "golang": {
         // ...
      }
   },
  "golang": {
    "suggest": {
      "autoTrigger": "trigger",
      // ...
    },
    "diagnostic": {
      // ...
    }
  }
}
```

Which configurations are allowed inside the language specific section? `suggest`, `diagnostic`, `coc.preferences`, all?

chemzqm:
The key need to be something like `"[javascript,typescript]"`

pedromss:
Makes sense if we want the same configs for multiple languages, what about which configurations can go in those? All of them?

chemzqm:
No, support some specific ones like enable options should be good enough

pedromss:
You mean `suggest.enable**` and `diagnostics.enable**` ?

PR #2507, opened 2020-10-23
add hover action target argument

smackesey:
The `doHover` action currently reads its target configuration directly from preferences. This PR changes it so that the target argument can optionally be provided as an argument (falling back to reading from preferences), following the style of the `jumpTo*` actions. This is useful because sometimes one wants to use different keybindings for a preview hover or floating hover, or to keep some hovered information open in the preview while using mostly using the float.

I updated the docs except for the Chinese text, because I don't know Chinese.

### DIFF
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -1819,8 +1819,9 @@ Acceptable {action} names for |CocAction()| and |CocActionAsync|.
 	Show documentation of the current word in a preview window.
 	Return `v:false` when hover not found.
 
-	{hoverTarget}: optional specification for where to show hover info, defaults to
-	`coc.preferences.hoverTarget` in `coc-settings.json`.
+	{hoverTarget}: optional specification for where to show hover info,
+	defaults to `coc.preferences.hoverTarget` in `coc-settings.json`.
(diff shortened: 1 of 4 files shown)


doc/coc.txt
```diff
@@ -1814,12 +1814,13 @@ Acceptable {action} names for |CocAction()| and |CocActionAsync|.
...
 
-"doHover"						*coc-action-doHover*
+"doHover" [{hoverTarget}]					*coc-action-doHover*
 
 	Show documentation of the current word in a preview window.
 	Return `v:false` when hover not found.
 
-	Use `coc.preferences.hoverTarget` to change hover behavior.
+	{hoverTarget}: optional specification for where to show hover info, defaults to
+	`coc.preferences.hoverTarget` in `coc-settings.json`.
```
chemzqm, line 1823: Please add acceptable values of `hoverTarget`

chemzqm (changes requested):

doc/coc.txt
smackesey: OK, I pushed a commit with those added.

PR #2518, opened 2020-10-26
Rework highlight for float window/popups

chemzqm:
Could work, but not finished yet.

- [x] Use syntax command for highlight, avoid background nvim instance.
- [x] Parse highlights and code blocks for markdown content.
- [x]  Simplify float window create by one vim call.

**Note** the API `FloatBuffer.setLines` is changed, extension author may need to change their code.


chemzqm:
@fannheyward @oblitum It's finished, consider try it out.

src/markdown/index.ts
```diff
@@ -0,0 +1,135 @@
...
+  let codes: CodeBlock[] = []
+  let currline = 0
+  let inCodeBlock = false
+  let filetype: string
+  let startLnum = 0
+  let parsed = marked(content)
+  for (let line of parsed.replace(/\s*$/, '').split(/\n/)) {
+    if (!line.length) {
+      let pre = lines[lines.length - 1]
+      if (pre.length) {
```
fannheyward, line 85: `pre` maybe undefined. reproduced with gopls

src/markdown/index.ts
```diff
@@ -0,0 +1,135 @@
...
+        startLnum = currline
+      } else {
+        inCodeBlock = false
+        codes.push({
+          filetype,
+          startLine: startLnum,
+          endLine: currline
+        })
+      }
+      if (pre.length) {
```
fannheyward, line 107: pre maybe undefined. reproduced with gopls

doc/coc.txt
```diff
@@ -2359,6 +2359,36 @@ Note: highlight commands should appear after the |:colorscheme| command or use
...
+  Default: `hi default link CocMarkdownCode     markdownCode`
+
+  The highlight for markdown code in floating window/popup.
+
+CocMarkdownHeader					*CocMarkdownHeader*
+
+  Default: `hi default link CocMarkdownHeader   markdownH1`
+
+  The highlight for markdown header in floating window/popup.
+
```
fannheyward, line 2391: `CocMarkdownLink` doc is missing

src/__tests__/modules/floatFactory.test.ts
```diff
@@ -29,19 +29,9 @@ describe('FloatFactory', () => {
...
       content: 'f'.repeat(81)
     }]
-    let called = false
-    let fn = (winid, bufnr) => {
-      called = true
-      expect(floatFactory.bufnr).toBe(bufnr)
-      expect(floatFactory.window.id).toBe(winid)
-    }
-    floatFactory.on('show', fn)
     await floatFactory.show(docs)
```
fannheyward, line 32: The doc hover can't be configured, do we need something like `signatureFloatMaxWidth` to set hover float width? Currently, the hover float width is hard-coded to 80.

fannheyward: I point the wrong file, I mean https://github.com/neoclide/coc.nvim/blob/876b8556e8a5610c07071b1c9ae67d987b74feeb/src/handler/index.ts#L1346

src/markdown/renderer.ts
```diff
@@ -0,0 +1,367 @@
...
+          .replace(/[^\w:]/g, '')
+          .toLowerCase()
+      } catch (e) {
+        return ''
+      }
+      if (prot.indexOf('javascript:') === 0) {
+        return ''
+      }
+    }
+    if (text && text != href) return styles.blue(text)
```
fannheyward, line 342: `[MDN Reference](https://developer.mozilla.org/docs/Web/HTML/Element/head)` is rendered to `MDN Reference`.
reproduced with coc-html

fannheyward:
```
    if (text === href) {
      return this.o.link(this.o.href(href))
    }

    return `[${text}](${href})`
```

how about this?

chemzqm: No, it's too long

fannheyward: 截屏2020-11-09 14 36 52

but the rendered link text is useless in vim

src/__tests__/modules/floatFactory.test.ts
chemzqm: It's not shown automatically, I don't think there's need to configure it.

fannheyward (approved):

fannheyward:
Tested with gopls/rust-analyzer/html/css/clangd, works great!

voldikss:
Seems `marked` uncorrectly parses quote with the `smartypants` option.

oblitum:
Agreed.

img-2020-11-09-120654

PR #2522, opened 2020-10-26
fix(ale): ensure refreshOnInsertMode is respected for Ale

abnerf:
The refreshOnInsertMode configuration flag was not being respected in case diagnostics were being set to the Ale plugin.

Apparently, this is a regression as this problem has been fixed in a previous PR: https://github.com/neoclide/coc.nvim/pull/1236/files


chemzqm:
It's expected, you need configure ale

abnerf:
> It's expected, you need configure ale

@chemzqm I am seeing exactly what was described on this PR before: https://github.com/neoclide/coc.nvim/pull/1236#issuecomment-537570151

Configuring Ale is not working and the results are the same as in this old PR. Curious to know why you mentioned this old PR was right and this new one (which is the same fix) is not. Did you find another way to configure Ale to fix it? Thanks!

PR #2525, opened 2020-10-27
feat: enable extensions auto-update

fannheyward:
We now have `coc.preferences.silentAutoupdate`, it's better to enable extensions auto-update by default: 

https://github.com/neoclide/coc.nvim/blob/fe165b2b62640b19523059a70a4accde35f39112/data/schema.json#L973-L977

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -848,7 +848,7 @@
     },
     "coc.preferences.extensionUpdateCheck": {
       "type": "string",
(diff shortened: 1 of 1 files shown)


fannheyward:
cc @chemzqm

chemzqm:
No, I don't think it's a good idea since the user would be frustrated when they got a broken extension update automaticlly.

Issue #2521, opened 2020-10-26
Make coc#float#scroll work in visual mode?


oblitum:
**Is your feature request related to a problem? Please describe.**

#2472 has added `coc#float#scroll` which works in normal and insert mode. Sadly (tested on nvim 0.4.4), it doesn't work on visual mode, which is an important use case on snippet expansion and jump placeholder.

**Describe the solution you'd like**

I wished it could be made to work as it works for insert and normal mode.

**Describe alternatives you've considered**

`coc#util#float_scroll_i`, which was added before #2472, can scroll in visual mode, the problem is that it was not improved to handle the scrollbar that was introduced by #2472, so even though it scrolls, the scrollbar is static on top, it never updates.

I'm not sure whether code from `coc#util#float_scroll_i` can be used as reference to make `coc#float#scroll` work in visual mode, or whether `coc#util#float_scroll_i` can be improved to update scrollbar position. I tried the latter by appending a call to `coc#float#nvim_scrollbar` to it but it didn't work.

**Additional context**

I commented about this [here](https://github.com/neoclide/coc.nvim/issues/2472#issuecomment-714681891), so I'm just opening this new issue so that won't get lost in comments.

chemzqm:
Nope, it can't be correct IMO

oblitum:
Oh, just managed to make `coc#util#float_scroll_it` to update the scrollbar state in visual mode, I just had to use the `timer_start` hack instead of calling `coc#float#nvim_scrollbar` directly:

```diff
diff --git a/autoload/coc/util.vim b/autoload/coc/util.vim
index 08ee758..9ead5e0 100644
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -115,6 +115,7 @@ function! coc#util#float_scroll_i(amount)
   endif
   call nvim_win_set_var(float, 'coc_float_scroll_last_amount', a:amount)
   call nvim_win_set_cursor(float, pos)
+  call timer_start(10, { -> coc#float#nvim_scrollbar(float)})
   return ''
 endfunction
```

oblitum:
My current mappings after that patch being applied locally:

```vim
let g:coc_snippet_next = '<c-l>'
let g:coc_snippet_prev = '<c-h>'
nnoremap <silent><expr> <c-j> coc#float#has_scroll() ? coc#float#scroll(1, 1) : "\<c-j>"
[...]

PR #2532, opened 2020-10-28
Add coc#float#nvim_scroll() with synchronized scrollbar

oblitum:
- Replace recent `coc#util#float_scroll_i` for something better named
- Add scrollbar synchronization
- Fix missing calculation of &scrolloff
- Nvim-only. Only useful for visual mode mappings (see added doc)

Partially solves #2521

### DIFF
--- a/autoload/coc/float.vim
+++ b/autoload/coc/float.vim
@@ -863,6 +863,49 @@ function! coc#float#nvim_check_related() abort
   endfor
 endfunction
 
+" Scroll float in any mode (neovim only)
+" Only really useful for visual mode scroll, where coc#float#scroll
+" is not yet implemented
+function! coc#float#nvim_scroll(forward, ...)
+  let float = coc#float#get_float_win()
+  if !float | return '' | endif
+  let buf = nvim_win_get_buf(float)
+  let buf_height = nvim_buf_line_count(buf)
+  let win_height = nvim_win_get_height(float)
+  if buf_height < win_height | return '' | endif
+  let pos = nvim_win_get_cursor(float)
+  let amount = (a:forward == 1 ? 1 : -1) * get(a:, 1, max([1, win_height/2]))
+  let scrolloff = &scrolloff*2 < win_height ? &scrolloff : 0
+  try
+    let last_amount = nvim_win_get_var(float, 'coc_float_scroll_last_amount')
+  catch
+    let last_amount = 0
+  endtry
+  if amount > 0
+    if pos[0] == 1
+      let pos[0] += amount + win_height - scrolloff*1 - 1
+    elseif last_amount > 0
+      let pos[0] += amount
+    else
+      let pos[0] += amount + win_height - scrolloff*2 - 1
+    endif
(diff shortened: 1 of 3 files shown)


chemzqm:
Please add documentation.

chemzqm:
You can use `redraw` instead of timer.

oblitum:
Hi @chemzqm, I see there's lots of refactoring/deprecation on util.vim. So I think it's best to also move this function to float.vim? Given that it's pretty recent and not many people are depending on it, what you think of simply moving it there without "deprecating" the not-very-old function?

So this is a list of questions to answer:

- Can I just move this function to float.vim without deprecation notice?
- Can I rename this function completely? `float_scroll_i` isn't a good name for functions in `coc#float`, should it be named `coc#float#something`, do you have an idea for a  name? Maybe `coc#float#nvim_scroll`?
- I've used the timer trick because calling it directly generated error E523 and the timer avoids that, it's not about a screen refresh problem. Does `redraw` fix that? Not sure how to use it and whether that suggestion is to fix that problem.
- I will add docs about the function to doc.txt and to this PR commit message. Is that sufficient?

To conclude, I think it's still fine for this function to exist in some form, without deprecation, because there's no other way to scroll in visual mode (I plan to add that to the docs).

chemzqm:
It's possible to cancel selection and make coc.nvim select the region again.

oblitum:
Okay. I'll rewrite this PR anyway. I won't cover rewriting it by a re-selectio mechanism. Sounds interesting though, but may present flicker side-effect.

autoload/coc/float.vim
```diff
@@ -863,6 +863,49 @@ function! coc#float#nvim_check_related() abort
...
+function! coc#float#nvim_scroll(forward, ...)
+  let float = coc#float#get_float_win()
+  if !float | return '' | endif
+  let buf = nvim_win_get_buf(float)
+  let buf_height = nvim_buf_line_count(buf)
+  let win_height = nvim_win_get_height(float)
+  if buf_height < win_height | return '' | endif
+  let pos = nvim_win_get_cursor(float)
+  let amount = (a:forward == 1 ? 1 : -1) * get(a:, 1, max([1, win_height/2]))
+  let scrolloff = &scrolloff*2 < win_height ? &scrolloff : 0
```
chemzqm, line 878: Should use `getwinvar()` for scrolloff option here

chemzqm (changes requested):

autoload/coc/float.vim
oblitum: Okay. I'll add but I don't exactly agree because it's a global option, so there's solely one value for it, always present.

chemzqm: It's not on latest vim and neovim 0.5.0, coc.nvim change it to 0 on neovim 0.5.0

oblitum: Oh, okay, so this option does not exist anymore?? Do you know if it's being simply removed or an alternative is being offered? Because it was somewhat useful for people.

oblitum: Ah okay, I interpreted it badly, it's has become window local.

PR #2538, opened 2020-10-30
feat(diagnostic): filter unused/deprecated diagnostics

fannheyward:
The configuration names came from https://github.com/microsoft/vscode/blob/2d8bbc6976748c4b0c5e84be541674b89b9d91e4/src/vs/monaco.d.ts#L4063-L4064

### DIFF
--- a/src/diagnostic/manager.ts
+++ b/src/diagnostic/manager.ts
@@ -266,25 +266,20 @@ export class DiagnosticManager implements Disposable {
     for (let collection of collections) {
       let items = collection.get(uri)
       if (!items) continue
-      if (level && level < DiagnosticSeverity.Hint) {
-        items = items.filter(s => s.severity == null || s.severity <= level)
-      }
-      if (!showUnused) {
-        items = items.filter(d => {
-          if (d.tags && d.tags.includes(DiagnosticTag.Unnecessary)) {
-            return false
-          }
-          return true
-        })
-      }
-      if (!showDeprecated) {
-        items = items.filter(d => {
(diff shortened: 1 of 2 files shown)


src/diagnostic/manager.ts
```diff
@@ -259,14 +261,30 @@ export class DiagnosticManager implements Disposable {
...
+        })
+      }
+      if (!showDeprecated) {
+        items = items.filter(d => {
+          if (d.tags && d.tags.includes(DiagnosticTag.Deprecated)) {
+            return false
+          }
+          return true
+        })
+      }
```
chemzqm, line 287: It's not necessary to use two filters.

chemzqm (changes requested):

src/diagnostic/manager.ts
```diff
@@ -266,25 +266,20 @@ export class DiagnosticManager implements Disposable {
...
-        items = items.filter(d => {
-          if (d.tags && d.tags.includes(DiagnosticTag.Deprecated)) {
-            return false
-          }
-          return true
-        })
-      }
+
+      items = items.filter(d => {
+        if (level && level < DiagnosticSeverity.Hint && d.severity && d.severity > level) {
```
chemzqm, line 271: You can use `d?.severity` nad `d?.tags` instead.

chemzqm (changes requested):

src/diagnostic/manager.ts
fannheyward: @chemzqm`d` is not optional, but `d.severity` maybe nil. `d.tags?.includes()` is ok.

PR #2539, opened 2020-10-31
fix: coc#float#get_config_pum input args may be floats

yatli:


chemzqm:
I've made a fix for it 7f27e1d1

yatli:
thank you very much!

PR #2540, opened 2020-10-31
Let user override default yarn executable name.

theg4sh:
The issue is that Fedora have package nodejs-yarn and it provides executable named `nodejs-yarn`.
In the source `yarn` executable is hardcoded.

That patch allow users to override name via global variable `g:coc_nvim_yarn`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,10 @@ Try these steps when you have problem with coc.nvim.
 - Checkout the log of coc.nvim by command `:CocOpenLog`.
 - When you have issues with the language server, it's recommended to [checkout
   the output](https://github.com/neoclide/coc.nvim/wiki/Debug-language-server#using-output-channel).
+- Fedora provides `yarn` tool also from nodejs-yarn package. If you install it this way, add
+  `let g:coc_nvim_yarn='nodejs-yarn'`
+  into the head of your vimrc.
+
 
 ## Feedback
 
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -17,6 +17,12 @@ elseif has('mac')
   let s:activate = 'activate'
 endif
 
+" Let user customize yarn executable name.
+" E.g. Fedora provides executable named as nodejs-yarn.
+if !exists('g:coc_nvim_yarn')
+  let g:coc_nvim_yarn='yarn'
+endif
+
 function! coc#util#has_preview()
   for i in range(1, winnr('$'))
     if getwinvar(i, '&previewwindow')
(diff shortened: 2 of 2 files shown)


oblitum:
If the user has to do that on vim config, why not simply create a symlink on `~/.local/bin` instead?

theg4sh:
Not all users have `~/.local/bin` in theirs PATH. Btw, it's not good practice to hardcode paths critical for work.

> If the user has to do that on vim config, why not simply create a symlink on `~/.local/bin` instead?

Not all users have `~/.local/bin` in theirs PATH. Btw, it's not good practice to hardcode paths critical for work.

oblitum:
When I stated `~/.local/bin`, it was just an example, to mean to put a symlink or related in your `$PATH`. I don't think it's wise to ask projects to have many variables due to how myriad systems deviate from the norm, it just adds complexity, while the user could simply fix their system, not only for an specific project like Coc, but maybe for many others that have the same expectation at the same time. IMO, this is system at fault, and it's simple for user to fix. I don't have final word so, let's see..

theg4sh:
> simply fix their system
> IMO, this is system at fault, and it's simple for user to fix.

Don't think so. Fedora and, e.g., Ubuntu from different universes. It's ok that binaries and packages have different names. It was that way as long as I remember this distributions. The point is not to _fix the system_, but to _setup the tool_. I as new user of Coc didn't expect the quest of digging the problem in sources while installing plugin. I think you making too big a deal out of this.

oblitum:
> I think you making too big a deal out of this.

I'm just exposing a view, no hard feelings please. It's either fine that a merge happens, I was simply expressing the idea of asking every tool out there to have this variable because system X doesn't follow the norm and having to set such variable on each of those tools instead of opting for a single fix that applies for all of them at once, kinda cumbersome.

oblitum:
FWIW, you should add a doc to coc.txt.

fannheyward (approved):

chemzqm:
I've reverted the commit, since we don't encourage to compile from source code.

theg4sh:
> I've reverted the commit, since we don't encourage user compile from source code.

huh? compile? I did just `dnf install nodejs-yarn`, there was no compiled sources. Plugin was installed via Vundle. I don't understood, what did you meant?

theg4sh:
@chemzqm ?

oblitum:
@theg4sh check master with better solution already. Avoid spamming.

theg4sh:
@oblitum have not seen relation between revert's comment and changes in master. Ok, that explains all, thanks a lot.

PR #2546, opened 2020-11-02
feat: use sign group

xuanduc987:
Coc tracked which signs it had placed to clear those later, but this list is frequently out of sync with what actually shown on screen.

By placing all of Coc sign in their own group, we could clear all of those signs with 1 function call.

checkSigns method is used as workaround for this problem thus is no longer needed.

I had using this patch for several weeks, with no visible problems found.


chemzqm:
We need support neovim 0.3.2 for that feature, but sign_place function not exits on that version.

xuanduc987:
Note: sign group added in vim 8.1.0658 or neovim 0.4.0

chemzqm (approved):

PR #2562, opened 2020-11-06
Fixes #2561 replace mv with fs-extra to prevent unlink errors on FreeBSD

### LINKED ISSUE #2561: Unable to CoCInstall on FreeBSD 12.1
Any coc plugin I install results in an unlink error on FreeBSD 12.1.

```vim
:CocInstall coc-json
```

```shell
Install finished      

coc-json EPERM: operation not permitted, unlink '/tmp/nvimsFT3FR/coc-json-CZ6oLp'
```

I traced the problem to the use of the `mv` node module in `src/model/installer.ts`. Pull request will be submitted shortly.

iamruinous:
Fixes #2561 which causes unlink errors on FreeBSD 12.1

Switch use of `mv` to `fs-extra` which also removes the need to promisify some of the other `fs` calls.


fannheyward (approved):

fannheyward:
Tested and works as expected.

截屏2020-11-06 12 46 48

cc @chemzqm how to fix this check?

PR #2585, opened 2020-11-10
feat(list): Add cancel command

piersy:
This PR adds `:CocListCancel` which acts like a counterpart to `:CocListResume` allowing the list to be closed when the list is not the current window.


chemzqm:
Why need this?

piersy:
> Why need this?

Hi @chemzqm. This is needed to be able to close the list when the list is not the current window.

When using CocNext/CocPrev to jump to list elements it's useful to be able to close the list without having to move to the list window.

PR #2639, opened 2020-11-22
Fix global promise patch

### LINKED ISSUE #2637: Global patch to promise prototype causes typescript errors in other packages
## Result from CocInfo

Coc functionality is not effected, CocInfo output is irrelevant for this issue.

## Describe the bug

`coc.nvim` patches promise prototype to add `logError` globally ([file](https://github.com/neoclide/coc.nvim/blob/master/src/util/extensions.ts)), this causes issues when developing coc plugins using typescript. In my example it is a thrid party code `postcss` which has a [line of code](https://github.com/postcss/postcss/blob/0d89106b73bf6ab6c0848c26cca29cbe874de7c8/lib/lazy-result.d.ts#L16) such as

```ts
class LazyResult implements Promise<Result> {}
``` 

## Reproduce the bug

I've created a [repository](https://github.com/antonk52/cocnvim-typescript-pollution-bug) to reproduce the issue.

When attempting to build using typescript you will see the following error

```
node_modules/postcss/lib/lazy-result.d.ts:16:22 - error TS2420: Class 'LazyResult' incorrectly implements interface 'Promise<Result>'.
  Property 'logError' is missing in type 'LazyResult' but required in type 'Promise<Result>'.

16 export default class LazyResult implements Promise<Result> {
                        ~~~~~~~~~~

  node_modules/coc.nvim/lib/util/extensions.d.ts:6:5
    6     logError(): void;
          ~~~~~~~~~~~~~~~~~
    'logError' is declared here.


Found 1 error.
```

A potential fix would be to avoid patching promise globally and to log errors where it is needed manually/adding a util promise wrapper for this purpose.

antonk52:
fixes #2637

I've used a [codemod](https://gist.github.com/antonk52/9bcc1a9ff6fcb54db947881646e959e3) to fix this issue.

### DIFF
--- a/src/attach.ts
+++ b/src/attach.ts
@@ -5,8 +5,8 @@ import events from './events'
 import Plugin from './plugin'
 import semver from 'semver'
 import { objectLiteral } from './util/is'
-import { URI } from 'vscode-uri';
-import logError from "./util/extensions";
+import { URI } from 'vscode-uri'
+import logError from "./util/extensions"
 const logger = require('./util/logger')('attach')
 const isTest = global.hasOwnProperty('__TEST__')
 
--- a/src/attach.ts
+++ b/src/attach.ts
@@ -5,15 +5,15 @@ import events from './events'
 import Plugin from './plugin'
 import semver from 'semver'
 import { objectLiteral } from './util/is'
-import './util/extensions'
-import { URI } from 'vscode-uri'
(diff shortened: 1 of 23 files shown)


chemzqm:
No, the prototype of Promise could be changed in standard js as well, maje your tsc not check thet file.

PR #2658, opened 2020-11-27
fix(channels): name with space can't be opened

fannheyward:
Problem: output channel with space in name can be created, but can't be opened.

For example, `workspace.createOutputChannel('channel name with space')`, when do `openBuffer`, we can't open by `tab drop output:///channle name with space`.

~Solution: just replace all space with `_` before create/show, and replace `_` to space for better list UI.~

### DIFF
--- a/src/model/outputChannel.ts
+++ b/src/model/outputChannel.ts
@@ -68,7 +68,8 @@ export default class BufferChannel implements OutputChannel {
     let { nvim } = this
     let winid = await nvim.call('win_getid')
     nvim.pauseNotification()
-    nvim.command(`tab drop output:///${this.name}`, true)
+    const escapedName = await nvim.call('fnameescape', this.name)
+    nvim.command(`tab drop output:///${escapedName}`, true)
     if (preserveFocus) {
       nvim.call('win_gotoid', [winid], true)
     }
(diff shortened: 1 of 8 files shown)


chemzqm:
No, don't replace space with `_`, escape it instead.

chemzqm (changes requested):
Escape space when open list buffer.

src/model/outputChannel.ts
```diff
@@ -68,7 +68,8 @@ export default class BufferChannel implements OutputChannel {
     let { nvim } = this
     let winid = await nvim.call('win_getid')
     nvim.pauseNotification()
-    nvim.command(`tab drop output:///${this.name}`, true)
+    const escapedName = await nvim.call('fnameescape', this.name)
+    nvim.command(`tab drop output:///${escapedName}`, true)
```
chemzqm, line 72: Need escape for `bufname` property and use bufrname here,  we could also throw error for invalid characters for channel name at https://github.com/neoclide/coc.nvim/blob/master/src/channels.ts#L39, I think that limit the characters to `[\w\s-]` is good enough.

chemzqm (changes requested):

chemzqm:
I've fixed it.

PR #2677, opened 2020-12-02
feat(plugin): use tabe for CocOpenLog

fannheyward:
~By default `CocOpenLog` will open log file by `edit` in current buffer, it's better to use `tabe` for log view.~

Update: default behavior will follow `coc.preferences.jumpCommand`, but I think it's better to use tab...


chemzqm:
It's just personal taste, someone doesn't like tab.

PR #2685, opened 2020-12-08
fix(github): use ubuntu-20.04

fannheyward:


fannheyward:
cc @chemzqm preview PR missed.

PR #2686, opened 2020-12-08
fix(diagnostic): make sure diagnostics not null

fannheyward:


chemzqm:
Undefined value should clear exists diagnostics, I've made a fix with 3d763824

PR #2693, opened 2020-12-09
feat(language-client): done notification with timeout

fannheyward:
Some notification is too short to display, and disappear in a flash, it's better to remove it with a timeout.

```
2020-12-09T11:58:45.236 ERROR (pid:91407) [language-client-progressPart] - --begin
2020-12-09T11:58:45.245 ERROR (pid:91407) [language-client-progressPart] - --end
```

### DIFF
--- a/src/language-client/progressPart.ts
+++ b/src/language-client/progressPart.ts
@@ -113,16 +113,15 @@ export class ProgressPart {
     if (message && this._progressTarget === 'statusline') {
       const statusBarItem = this.statusBarItem
       statusBarItem.text = `${this.title} ${message}`
-      setTimeout(() => {
-        this.cancel()
-      }, 300)
     } else {
       if (this._resolve) {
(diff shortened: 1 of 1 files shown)


chemzqm:
There is timeout for progress module, so I think it's not need here.

fannheyward:
Found the timeout in progress module, but there was another problem: progress window can't be removed with quickly document change.

With this patch, the progress window can be removed.

fannheyward:
Steps to reproduce:

1. setup `rust-analyzer` language server
2. sample code:
```rust
fn main() {
    println!("{}", 1000000);
}
```
3. `x` on 0, will do indexing, progress notification appears and removed
4. `x` on 0 multiple times and quickly, multiple indexing progress windows appear, but not all be removed

chemzqm:
It's because promise of withProgress not create, I've fix that issue.

PR #2730, opened 2020-12-18
fix(completion): checkTrigger for triggerSources

### LINKED ISSUE #2728: Disabling suggest.autoTrigger no longer works
## Result from CocInfo

```
## versions

vim version: NVIM v0.5.0-844-g8c4648421
node version: v14.8.0
coc.nvim version: 0.0.79
coc.nvim directory: /home/idanarye/.vim/plugins/coc.nvim
term: xterm-kitty
platform: linux

## Log of coc.nvim

2020-12-18T00:14:23.261 INFO (pid:15809) [plugin] - coc.nvim 0.0.79 initialized with node: v14.8.0 after 25ms
2020-12-18T00:14:25.361 INFO (pid:15809) [completion-complete] - Results from: file
2020-12-18T00:14:29.220 INFO (pid:15809) [completion-complete] - Results from: file
2020-12-18T00:14:46.087 INFO (pid:15809) [completion-complete] - Results from: file


```

## Describe the bug

Configuring `suggest.autoTrigger` to `none` no longer works - the completion popup still gets opened automatically after typing certain characters like `.` or `/`.

I have bisected the commits and found that this bug was introduced in 365c63e12f82675dfe3ec485a57eb4af964e421d. When I checkout the commit before it `suggest.autoTrigger` works and I do not get uninvited completion popups.

## Reproduce the bug

- Create file `mini.vim` with：

```vim
set nocompatible
filetype plugin indent on
syntax on
set hidden

call coc#config('suggest.autoTrigger', 'none')
```

- Start (neo)vim with command: `vim -u mini.vim`

- Hit `i` to enter insert mode.
- Hit `/` (the bug still happens with other completion triggers, but `/` is the easiest to trigger with a minimal setup)
- A path completion popup is opened. It shouldn't open, because `mini.vim` sets `suggest.autoTrigger` to `none`.

fannheyward:
The point of 365c63e optimization is get trigger sources first before triggerCompletion, then disable checkTrigger to improve perf, did I understand correctly? @chemzqm . But source trigger checking doesn't check suggest.autoTrigger, this causes the issue.

This is a quick fix to #2728, maybe we need to refactor trigger  completions checking, include autoTrigger checking, sources trigger checking, coc_suggest_disable, blacklist etc.

close #2728

### DIFF
--- a/src/completion/index.ts
+++ b/src/completion/index.ts
@@ -314,7 +314,7 @@ export class Completion implements Disposable {
       if (!latestInsertChar) return
       let triggerSources = sources.getTriggerSources(pretext, doc.filetype)
       if (triggerSources.length) {
-        await this.triggerCompletion(doc, this.pretext, false)
+        await this.triggerCompletion(doc, this.pretext)
         return
       }
       this.triggerTimer = setTimeout(async () => {


chemzqm:
Will fix it.

PR #2732, opened 2020-12-18
feat(typings): add Diagnostic namespace

fannheyward:


fannheyward:
@chemzqm is this the correct way to add typings?

chemzqm:
You have to export `Diagnostic` in `src/index.ts`, otherwise it won't work.

PR #2741, opened 2020-12-20
refactor resolveRoot

wwwjfy:
This commit uses the "shift-right" way to find workspace folders.

Earlier version starts from root "/" until the path of the current file,
stopping at where the root patterns are found. The new version starts
from the current file and traverse upwards.

It aims to fix the situation of "embeded" workspaces usually seen in
monorepo. For example, a monorepo root has a ServiceA folder,
"ServiceRoot/ServiceA", both of which have files matching root patterns.
The expected workspace folder should be ServiceRoot/ServiceA, but not
ServiceRoot.

This issue is triggered when jumping from other workspace folder like
ServiceRoot/ServiceB.

Another example is using vim-plug which clones coc.nvim in
~/.vim/plugged folder. It resolves to ~/.vim which contains .git, while the
expected result is ~/.vim/plugged/coc.nvim.

### DIFF
--- a/src/__tests__/modules/util.test.ts
+++ b/src/__tests__/modules/util.test.ts
@@ -159,7 +159,7 @@ describe('resolveRoot', () => {
   it('should resolve from parent folders', () => {
     let root = path.resolve(__dirname, '../extensions/snippet-sample')
     let res = resolveRoot(root, ['package.json'])
-    expect(res.endsWith('coc.nvim')).toBe(true)
+    expect(res.endsWith('extensions')).toBe(true)
   })
 
   it('should not resolve to home', () => {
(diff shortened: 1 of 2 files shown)


chemzqm:
It works like by design, check wiki.

chemzqm:
We may change it in the future, not for now.

wwwjfy:
Is this the wiki you referred to? https://github.com/neoclide/coc.nvim/wiki/Using-workspaceFolders

> Note workspace folder is resolved from up to down, cwd is checked first when the file is inside cwd,.

It's a bit vague and doesn't tell the reason.

The situation I met is I work in a hugo Go monorepo, and I do open vim in workspace root, but when I jump to an imported package which resides in another subfolder, other than my workspace root. The monorepo is added to be the workspace folders, which takes forever and infinite memory for gopls to index.

I understand there may be other cases we design it this way. would be appreciate if you can tell the reason.
I'll just use my fork then.

chemzqm:
Because some language servers needs the folder to be project root, search from filepath won't work, add option might help.

wwwjfy:
Thanks. I'll try to add an option

wwwjfy:
I tried to add an option [here](https://github.com/wwwjfy/coc.nvim/commit/e20822a2b94e9e8bbe5719a28dc7e1c157f60b9d)
It's not reflected in this PR as it's closed.

Let me know if I miss anything. Thanks~

wwwjfy:
I'll create another PR later.

PR #2742, opened 2020-12-20
fix: gotoDefinition workaround for java on Windows

AdnoC:
Fixes: https://github.com/neoclide/coc-java/issues/82

Adds specific workarounds to fix "go to definition" functionality on Windows.
Two workarounds are implemented:

As the `:edit` command was failing, we instead call `nvim_buf_set_name` to set the buffer name without expansion. Then we `:edit` the current file in order to run autocommands.

Additionally, the URI of the loaded definition was `unknown:2`, which is incorrect. So we do not normalize the path when in a `jdt://` file.

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -157,8 +157,15 @@ function! coc#util#jump(cmd, filepath, ...) abort
   if (has('win32unix'))
     let path = substitute(a:filepath, '\v\\', '/', 'g')
   endif
-  let file = fnamemodify(path, ":~:.")
-  exe a:cmd.' '.fnameescape(file)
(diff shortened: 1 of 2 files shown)


chemzqm:
`nvim_buf_set_name` exists on neovim only.

AdnoC:
You are correct.

chemzqm:
I'll fix the issues while avoid `nvim_buf_set_name`

AdnoC:
https://github.com/neoclide/coc.nvim/commit/bf947691c214deaf05d3e2dc82ff6d51b9121c17 does not fix the issue. Pulled master, rebuilt, and am still getting the same error.

PR #2760, opened 2020-12-23
refactor resolveRoot

wwwjfy:
This commit uses the "shift-right" way to find workspace folders.

Earlier version starts from root "/" until the path of the current file,
stopping at where the root patterns are found. The new version starts
from the current file and traverse upwards.

It aims to fix the situation of "embeded" workspaces usually seen in
monorepo. For example, a monorepo root has a ServiceA folder,
"ServiceRoot/ServiceA", both of which have files matching root patterns.
The expected workspace folder should be ServiceRoot/ServiceA, but not
ServiceRoot.

This issue is triggered when jumping from other workspace folder like
ServiceRoot/ServiceB.

Another example is using vim-plug which clones coc.nvim in
~/.vim/plugged folder. It resolves to ~/.vim which contains .git, while the
expected result is ~/.vim/plugged/coc.nvim.

This is a followup of #2741, adding an option to be language specific.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -25,6 +25,11 @@
           "default": false,
           "description": "Disable workspaceFolders feature for this languageserver."
         },
+        "workspaceFolderBottomUpResolution": {
+          "type": "boolean",
+          "default": false,
+          "description": "Find the workspace folder in the bottom-to-up manner."
+        },
         "disableSnippetCompletion": {
           "type": "boolean",
           "default": false,
--- a/data/schema.json
+++ b/data/schema.json
@@ -25,11 +25,6 @@
           "default": false,
           "description": "Disable workspaceFolders feature for this languageserver."
         },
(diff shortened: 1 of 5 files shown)


data/schema.json
```diff
@@ -25,6 +25,11 @@
           "default": false,
           "description": "Disable workspaceFolders feature for this languageserver."
         },
+        "workspaceFolderBottomUpResolution": {
```
chemzqm, line 28: It should not belongs be language server configuration, since user might using extension of coc.nvim,
use `workspace.workspaceFolderBottomUpFiletypes` would be better.

chemzqm (changes requested):

wwwjfy:
Thanks for the comment. Fixed.
Please suggest if I need to update documentation in the same PR.

PR #2766, opened 2020-12-24
feat(util): add support regex for language server's rootPattern

quanlou:
This PR aim to tackle same issue mentioned at #2741 about setting workspace at the root of monorepo which I also experienced myself. My usage is gopls (the Go language server), that letting gopls index our entire monorepo is not possible and I might arbitrary assume that this issue should not in particular for gopls but for any language server at big monorepo size

By supporting regex at languageserver's rootPattern config, it'll leave a choice for user to "pin" their workspace to a certain format

### DIFF
--- a/src/util/fs.ts
+++ b/src/util/fs.ts
@@ -56,10 +56,23 @@ export async function isGitIgnored(fullpath: string): Promise<boolean> {
   return false
 }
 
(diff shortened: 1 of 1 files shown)


chemzqm:
The pattern should be file inside folder, but you're using it match folder, it's not a good solution, resolve from bottom to top should cover your case, checkout #2760

PR #2786, opened 2021-01-02
use empty line to delimit hover contents for preview target 

tomtomjhj:
`"coc.preferences.hoverTarget": "preview"` uses vim's markdown syntax
plugin, which usually highlights

<pre>
text
---
</pre>

as a markdown header. ~~To prevent this, insert a newline before '---'.~~

**Before**:
image

**After**:

image

## Note
* ~~This somehow doesn't add the newline for `"coc.preferences.hoverTarget": "float"`, so it won't disturb users using that option.~~
* Ideally, it'd be great if `"coc.preferences.hoverTarget": "preview"` can make use of the markdown renderer that `"float"` is using.

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -811,7 +811,7 @@ export default class Handler {
     }
     let lines = docs.reduce((p, c) => {
       let arr = c.content.split(/\r?\n/)
-      if (p.length > 0) p.push('---')
+      if (p.length > 0) p.push('')
       p.push(...arr)
       return p
     }, [])


chemzqm:
Why not just make it empty line?

tomtomjhj:
Do you mean `if (p.length > 0) p.push('')`? Hmm that makes sense.

tomtomjhj:
Oh what about `if (p.length > 0) p.push('———')` (em dash)

chemzqm:
I prefer empty line

PR #2803, opened 2021-01-10
add handle floatActions

hanh090:
Give floatActions setting. If false, using quickpick


chemzqm:
Thanks, missing that part.

PR #2807, opened 2021-01-11
feat(workspace): add option to disable cwd check when resolving root directory

kiancross:
### Why?

When resolving the root directory, the current working directory is checked before doing a top-down search:

https://github.com/neoclide/coc.nvim/blob/f3b4a47780829f663740f1d276879060b45e8923/src/util/fs.ts#L63

In some cases, it is preferable to skip this check and *always* do a top-down search, particularly in some mono-repo setups.

### How?

This PR adds the option `workspace.workspaceFolderCheckCwd`, which defaults to `true`, so the default behaviour is unchanged. When this option is set to `false`, the current working directory is not checked.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -800,6 +800,11 @@
         "type": "string"
       }
     },
+    "workspace.checkCwd": {
+      "type": "boolean",
+      "default": true,
+      "description": "Whether the cwd directory should be checked first when resolving the root directory"
+    },
     "list.indicator": {
       "type": "string",
       "default": ">",
--- a/data/schema.json
+++ b/data/schema.json
@@ -800,7 +800,7 @@
         "type": "string"
       }
     },
-    "workspace.checkCwd": {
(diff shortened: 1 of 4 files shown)


kiancross:
Looks like the checks are failing due to neovim/neovim#13733.

data/schema.json
```diff
@@ -800,6 +800,11 @@
         "type": "string"
       }
     },
+    "workspace.checkCwd": {
```
chemzqm, line 803: Please use `workspace.workspaceFolderCheckCwd`

chemzqm (changes requested):

data/schema.json
kiancross: @chemzqm OK I have made this change.

chemzqm (approved):

kiancross:
@chemzqm It looks like some of the tests are timing out? They started failing after I merged master into this branch...

PR #2815, opened 2021-01-14
Improvements for prompt win in vim

xuanduc987:
- `termwinkey` default to <kbd>C-W</kbd> in Vim, thus we must press <kbd>C-W .</kbd> to delete word -> set `termwinkey` to <kbd>C-\\</kbd> in prompt win buffer
- <kbd>ESC</kbd> is not recognized immediately in prompt win -> set `escapeCodeTimeout=0`

### DIFF
--- a/bin/prompt.js
+++ b/bin/prompt.js
@@ -2,11 +2,10 @@
  * Used for prompt popup on vim
  */
 const readline = require("readline")
-readline.emitKeypressEvents(process.stdin)
-process.stdin.setRawMode(true)
 const rl = readline.createInterface({
   input: process.stdin,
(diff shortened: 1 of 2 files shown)


bin/prompt.js
```diff
@@ -2,11 +2,10 @@
  * Used for prompt popup on vim
  */
 const readline = require("readline")
-readline.emitKeypressEvents(process.stdin)
-process.stdin.setRawMode(true)
```
xuanduc987, line 6: `readline.emitKeypressEvents(process.stdin)` is not necessary because `readline.createInterface` implicit called it

PR #2825, opened 2021-01-16
Use window height to set topline for list previews

lanza:
For splitRight == true splits this was setting the selected line to line
number 3 and leaving (for my laptop) 62 lines below unused with only 2
lines of context above. Make this a bit more even by splitting to the
middle.


chemzqm:
It's by design, we could provide configuration for that.

PR #2831, opened 2021-01-18
feat: CocListClose event

weirongxu:
I use `CocList name` in coc-explorer, but in the floating window, I need to rerender explorer when the `CocList` is closed, so I need something like this.
I named it `CocListClose`, if there is a better name or solution, I can adjust it.

### DIFF
--- a/src/list/session.ts
+++ b/src/list/session.ts
@@ -304,7 +304,10 @@ export default class ListSession {
         await wait(10)
       }
     }
+    nvim.pauseNotification()
     nvim.call('coc#prompt#stop_prompt', ['list'], true)
+    nvim.call('coc#util#do_autocmd', ['CocListClose'], true)
+    await nvim.resumeNotification()
   }
 
   public toggleMode(): void {


chemzqm:
You have to save `winid` of list and listen to `BufWinLeave` event from `events` module.

weirongxu:
@chemzqm I do this now, but it's too hack and buggy. Once coc.nvim is updated, bugs will occur

https://github.com/weirongxu/coc-explorer/blob/master/src/source/source.ts#L264-L269

weirongxu:
I will try to detect list from `filetype`, maybe it can implement this feature more stably

chemzqm:
Could you explain why you need rerender explorer on list window close?
Does it need rerender on window layout change?

chemzqm:
```
    const shownExplorerEmitter = new Emitter<void>();
    const listDisposable = listManager.registerList(list);
    // @ts-ignore TODO
    await listManager.start([list.name]);
```
CocList is not designer for temporary use case, looks like you should use `window.showMenuPicker` instead.

weirongxu:
The first video is that I want to search for files from a directory in explorer then reveal and expend to the file node

https://user-images.githubusercontent.com/1709861/104890884-26ca3d80-59ab-11eb-911d-4d263eaef64c.mp4

This one is that I select the drives list from the `CocList`

https://user-images.githubusercontent.com/1709861/104890893-292c9780-59ab-11eb-8a55-bd17b09d3eb8.mp4

This is I use `CocList` to select the coc-explorer actions
https://user-images.githubusercontent.com/1709861/104891608-22eaeb00-59ac-11eb-8119-490f9207a554.mp4

I will try the `window.showMenuPicker`, but if the menu is too long, it maybe not support this requirement very well.

chemzqm:
You should avoid things like
```ts
    fileList.showHidden = this.showHidden;
    fileList.rootPath = path;
    fileList.recursive = recursive;
    fileList.revealCallback = async (loc) => {
      await task.waitShow();
      const [, notifiers] = await this.revealNodeByPathNotifier(
        Uri.parse(loc.uri).fsPath,
      );
      await Notifier.runAll(notifiers);
    };

    const task = await this.startCocList(fileList);
```

Register the list once, then use
```ts
nvim.command(`CocList explorerFiles --rootPath=/a/c -R -I -H`, true)
```
to invoke the list with arguments instead.

weirongxu:
Thank you so much, I will refactor the related code and use command arguments instead of class properties

PR #2872, opened 2021-02-05
Always apply code action if one specific action is requested

### LINKED ISSUE #2871: CocAction with specific action to run shows floating window
## Result from CocInfo

```
## versions

vim version: NVIM v0.4.4
node version: v14.15.0
coc.nvim version: 0.0.80-6e5a2aaeb5
coc.nvim directory: redacted
term: iTerm.app
platform: darwin
```

## Describe the bug

The very nice new floating window shows up even when I call `CocAction` with the specific name of a single action.
So now instead of my keybinding executing the action it triggers this window and then does not directly execute the action.

How can I tell the `CocAction` function to not show the floating window?

Example usage:
```vim
function! s:CodeActionFromSelected(type)
  call CocAction('codeAction', a:type, g:nextCodeAction)
endfunction

function! s:RunCodeAction(action)
  let g:nextCodeAction=a:action
  set operatorfunc=s:CodeActionFromSelected
  normal g@iw
endfunction

nnoremap <buffer> <silent> <leader>fww :call <SID>RunCodeAction('Wrap with Container')<cr>viw
```

Kavantix:
Fixes #2871 

@chemzqm Let me know if this fix would be acceptable or if you would want something more elaborate like adding a parameter to configure this (or a preference maybe)

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -448,6 +448,10 @@ export default class Handler {
     let codeActions = await this.getCodeActions(doc, range, Array.isArray(only) ? only : null)
     if (only && typeof only == 'string') {
       codeActions = codeActions.filter(o => o.title == only || (o.command && o.command.title == only))
+      if (codeActions.length == 1) {
+        await this.applyCodeAction(codeActions[0])
+        return
+      }
     }
     if (!codeActions || codeActions.length == 0) {
       window.showMessage(`No${only ? ' ' + only : ''} code action available`, 'warning')


fannheyward:
If there is only one codeAction, this patch will make it runs automatically. This may cause some unexpected behavior, for example:

截屏2021-02-06 18 05 54

With this patch, this `Move to a new file` codeAction will be applied.

Kavantix:
It will only do that if CocAction is called with the only parameter and if that only parameter is a string.
So I thought if someone does that then the intention is to run that specific action so there is no need to show it.

In the version of coc before the floating windows my code actions were also applied automatically if I passed a single only param as string

Also @fannheyward, do you perhaps have a suggestion how it should work since the auto apply is the whole point of my issue and pr

PR #2905, opened 2021-02-22
fix(workspace): fire BufReadCmd for scheme:/

fannheyward:
`deno lsp` returns `deno:/xxx` which is handled by VSCode, https://github.com/denoland/deno/issues/9514, coc only supports `deno://xxx`.

@chemzqm  I don't find any _stand_ format in LSP, so you can close if this may cause other problems.


chemzqm:
It should use `scheme:/` here which could handle `scheme:///` as well.

PR #2914, opened 2021-02-24
feat(handler): coc-codeaction-cursor

### LINKED ISSUE #2913: codeAction at current position
@chemzqm I think this is still an open problem in CoC. (Note that I'm not a CoC user, but a rust-analyzer maintainer.) I think providing special commands for running code action on the whole buffer or the current line doesn't make sense; and on the other hand, it seems there is no way of running code actions with an empty range?

My impression is that there's a different interpretation of what the range in the code action request means at play here. Getting code actions with a certain range is not the same thing as getting all possible code actions for that range (/ that can affect that range). At least rust-analyzer does not interpret it that way, and instead chooses to use the range basically as an input to the code actions (so we'll return actions that will do something with the given range, e.g. extract it to a new variable). I think the spec is unclear on this, but VSCode uses it that way (sending the current selection as the range, or an empty range if there's no selection), and Emacs does the same. rust-analyzer will send lots of code actions if you provide it an empty range at a certain point.

_Originally posted by @flodiebold in https://github.com/neoclide/coc.nvim/issues/1981#issuecomment-783350621_

fannheyward:
close #2913

### DIFF
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -1,6 +1,6 @@
 *coc-nvim.txt*				LSP support for Vim & Neovim.
 
-Version: 0.0.79
+Version: 0.0.80
 Author: Qiming Zhao <chemzqm at gmail.com>
 License: MIT license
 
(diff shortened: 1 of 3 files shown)


chemzqm:
Need docs.

chemzqm (approved):

PR #2915, opened 2021-02-24
fix(extensions): activate extension on command

fannheyward:
Command of `onCommand:xxx` in `activationEvents` should be listed in `:CocList commands` and active the extension.

This bug was brought in by https://github.com/neoclide/coc.nvim/pull/2295 , should fix https://github.com/iamcco/coc-flutter/issues/63 cc @iamcco


iamcco:
Thanks

PR #2969, opened 2021-03-13
Recreate disposed channels

Kavantix:
Currently if an extension disposes an output channel for any reason a new one cannot be created again with the same id.
This PR fixes that

### DIFF
--- a/src/channels.ts
+++ b/src/channels.ts
@@ -39,7 +39,7 @@ export class Channels {
   public create(name: string, nvim: Neovim): OutputChannel | null {
     if (outputChannels.has(name)) {
       const channel = outputChannels.get(name)
-      if (!channel.disposed) return channel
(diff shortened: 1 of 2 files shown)


src/channels.ts
```diff
@@ -37,7 +37,10 @@ export class Channels {
   }
 
   public create(name: string, nvim: Neovim): OutputChannel | null {
-    if (outputChannels.has(name)) return outputChannels.get(name)
+    if (outputChannels.has(name)) {
+      const channel = outputChannels.get(name)
+      if (!(channel as any).disposed) return channel
```
Kavantix, line 42: I'm doing an as any here because I don't want to expose the `disposed` getter in the types

chemzqm:
I've made another fix: 98a0c6db

Issue #2984, opened 2021-03-22
Configurate doHover floating window size


ytian81:
I am trying to use the following to configurate the hover float window but the window size doesn't seem to be able to change even after `CocRebuild` and `CocRestart`
```
    "hover.floatMaxWidth": 200,
    "hover.floatMaxHeight": 200,

```

_Originally posted by @ytian81 in https://github.com/neoclide/coc.nvim/issues/2979#issuecomment-804443305_

Avi-D-coder:
Can reproduce.
My hover float is now 80 characters wide, the setting is ignored.

chemzqm:
It's limited to 80.

ytian81:
Is it possible to set the width from 80 to some other values?

chemzqm:
Change this line https://github.com/neoclide/coc.nvim/blob/master/autoload/coc/float.vim#L849

ytian81:
It would be ideal to set this value to some large value like 500 or 1000 and then use `"hover.floatMaxWidth": 80` to limit the width instead.

chemzqm:
Fixed by cc1ad566

PR #2986, opened 2021-03-23
Change the hover floating window default max width

ytian81:
This PR changes the hover floating window max width from 80 to 500. Note the window width is still configurable from coc-settings, as the value changed by this PR is only a upper bound of the setting value.

This PR would resolve https://github.com/neoclide/coc.nvim/issues/2984.


chemzqm:
I've made fix needed.

PR #2989, opened 2021-03-25
feat(float): do not conceal in the floats

whisperity:
Several Vim themes are broken with regards to concealments, especially in coc's popup. In many cases, user-defined concealments are hard to use or unwanted in the context of writing code.

While normally you can set their `conceallevel` manually as you wish, coc's popup/float feature **hardwires it** to `2`.

Here's a screenshot with the [`iceberg`](http://github.com/cocopon/iceberg.vim) theme. Several things, such as the `⛏️`, the bold  **`B`** for `bool`, the `.` instead of `::` in `llvm::cl`, or `⊥`  instead of `false` come from conceals.
Note how `☥` and `∅` in the normal code (behind the popup) looks okay.

Screenshot_20210325_170607

As suggested, we will instead hardwire to `0`, resulting in no concealments:
Screenshot_20210325_171606

### DIFF
--- a/autoload/coc/float.vim
+++ b/autoload/coc/float.vim
@@ -168,7 +168,7 @@ function! coc#float#create_float_win(winid, bufnr, config) abort
   call setwinvar(winid, '&colorcolumn', 0)
   call setwinvar(winid, '&wrap', 1)
   call setwinvar(winid, '&linebreak', 1)
-  call setwinvar(winid, '&conceallevel', 2)
+  call setwinvar(winid, '&conceallevel', 0)
   let g:coc_last_float_win = winid
   call coc#util#do_autocmd('CocOpenFloat')
   return [winid, bufnr]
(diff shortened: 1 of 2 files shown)


chemzqm:
My suggestion is simply disable `conceallevel` since it's not needed any more.

doc/coc.txt
```diff
@@ -1553,7 +1553,13 @@ g:coc_markdown_disabled_languages  			*g:coc_markdown_disabled_languages*
 	Filetype list that should be disabled for highlight in markdown block, ex: >
 
 	let g:coc_markdown_disabled_languages = ['html']
->
+
+g:coc_popup_conceal_disable 				*g:coc_popup_conceal_disable*
```
chemzqm, line 1557: No need to add g:coc_popup_conceal_disable support

chemzqm (changes requested):

whisperity:
@chemzqm 
> My suggestion is simply disable `conceallevel` since it's not needed any more.

Sounds good to me.

PR #3067, opened 2021-05-03
Limit get_float_win_list's output to coc.nvim windows

dstein64:
(I had originally opened PR #3066 with this proposed change, but the PR was automatically closed when I tried changing the base branch from `release` to `master`)

In https://github.com/dstein64/nvim-scrollview/issues/39, it was reported that `coc.nvim` functionality was not working properly after a recent update I made to [`nvim-scrollview`](https://github.com/dstein64/nvim-scrollview).

It appears that the scrolling functionality in `coc.nvim` considers all floating windows, instead of only those created by `coc.nvim`.

This PR updates the `get_float_win_list` function to only return floating windows that have the `float` window variable set, which matches how `has_float` and `close_all` functions check for windows created by `coc.nvim`. Along with this update, the `close_all` function was updated to exclude the aforementioned check for the `float` variable, since it's no longer necessary with `get_float_win_list` only returning windows with that variable set.

### DIFF
--- a/autoload/coc/float.vim
+++ b/autoload/coc/float.vim
@@ -30,9 +30,7 @@ function! coc#float#close_all() abort
   let winids = coc#float#get_float_win_list()
   for id in winids
     try
-      if getwinvar(id, 'float')
-        call coc#float#close(id)
-      endif
+      call coc#float#close(id)
     catch /E5555:/
       " ignore
     endtry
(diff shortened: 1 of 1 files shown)


dstein64:
I just saw in [CONTRIBUTING.md](https://github.com/neoclide/coc.nvim/blob/89a2dc8fc80c0834b07677b858fb8e6977fd663f/CONTRIBUTING.md) that there is a format request for commit messages. I'll re-open this PR following the conventions.

PR #3100, opened 2021-05-20
feat(markdown): ignore image in markdown doc

fannheyward:
ignore `alt` in markdown text.

### DIFF
--- a/src/markdown/index.ts
+++ b/src/markdown/index.ts
@@ -3,6 +3,7 @@ import Renderer from './renderer'
 import { parseAnsiHighlights } from '../util/ansiparse'
 import { Documentation } from '../types'
 import { byteLength } from '../util/string'
+import workspace from '../workspace'
 export const diagnosticFiletypes = ['Error', 'Warning', 'Info', 'Hint']
 const logger = require('../util/logger')('markdown-index')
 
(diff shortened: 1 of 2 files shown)


chemzqm:
Why exclude image links?

fannheyward:
截屏2021-05-20 下午2 55 57

fannheyward:
Maybe we should ignore base64 image url only?

chemzqm:
We need to use option to trim image url since the terminal could render image and vim may provide api to render image.

src/markdown/index.ts
```diff
@@ -140,6 +140,12 @@ export function parseMarkdown(content: string): DocumentInfo {
       }
       continue
     }
+
+    if (/!\[.*?\]\(.*?\)/.test(line)) {
```
chemzqm, line 144: The image may follow by other text, this would ignore them as well.

chemzqm: It should only be ignored when whole line is images, replace images with empty string for other cases.

chemzqm (changes requested):

src/markdown/index.ts
fannheyward: Replace image pattern to empty string and check  line.length, only ignore whole line empty.

gk488 (approved):

PR #3156, opened 2021-06-22
Activated code actions resolution

adamansky:
Since LSP v3.16.0 languages servers may implement resolution of choosen code actions.
Currently coc.nvim engaged `edit` code actions to be not resolved by default: https://github.com/neoclide/coc.nvim/blob/master/src/language-client/client.ts#L2375
But at the same time resolving of actions is not fully implemented.

For example simple code actions like `Import Qualifier (...)` don't work in `coc-java`.

I've tried to fix it, so please take a look.

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -580,7 +580,7 @@ export default class Handler {
   public async applyCodeAction(action: CodeAction): Promise<void> {
     if (action.data !== undefined) {
       return this.resolveCodeAction(action).then(resolved => {
-        if (resolved && resolved.data === undefined) { // Code action was resolved
+        if (resolved.data === undefined) { // Code action was resolved
           return this.applyCodeAction(resolved)
         }
       })
@@ -609,8 +609,8 @@ export default class Handler {
     }
   }
 
-  public async resolveCodeAction(action: CodeAction): Promise<CodeAction | null> {
-     let { doc } = await this.getCurrentState()
+  public async resolveCodeAction(action: CodeAction): Promise<CodeAction> {
+     const { doc } = await this.getCurrentState()
      return this.withRequestToken('codeaction resolve', token => {
       return languages.resolveCodeAction(doc.textDocument, action, token)
      }, true)
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -600,6 +600,20 @@ export default class Handler {
         }
       }
     }
+    if (action.data !== undefined) {
+      return this.resolveCodeAction(action).then(resolved => {
+        if (resolved && resolved.data === undefined) { // Code action was resolved
+          return this.applyCodeAction(resolved)
+        }
+      })
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


src/provider/codeActionmanager.ts
```diff
@@ -23,6 +23,22 @@ export default class CodeActionManager extends Manager<CodeActionProvider> imple
     })
   }
 
+  public async resolveCodeAction(document: TextDocument, action: CodeAction, token: CancellationToken): Promise<CodeAction | null> {
+    if (action.data === undefined) { // Code action doesn't need a resolution
+      return action
+    }
+    const provider = this.getProviders(document).map(p => p.provider).find(p => typeof p.resolveCodeAction === 'function')
+    if (provider !== undefined) {
```
fannheyward, line 31: 1. if no provider is found, we should return `action` or null?
2. return first if no provider found

fannheyward (changes requested):

src/provider/codeActionmanager.ts
adamansky: Okay! Fixed it

chemzqm:
* You should not check `action.data` to determine resolve or not.
* The action should contains `clientId`, no need for document

I will make a fix for it.

src/handler/index.ts
```diff
@@ -578,6 +578,13 @@ export default class Handler {
   }
 
   public async applyCodeAction(action: CodeAction): Promise<void> {
+    if (action.data !== undefined) {
+      return this.resolveCodeAction(action).then(resolved => {
```
fannheyward, line 582: it's better to use `await` here.

src/provider/codeActionmanager.ts
```diff
@@ -23,6 +23,22 @@ export default class CodeActionManager extends Manager<CodeActionProvider> imple
...
+  public async resolveCodeAction(document: TextDocument, action: CodeAction, token: CancellationToken): Promise<CodeAction> {
+    if (action.data === undefined) { // Action needs a resolution?
+      return action
+    }
+    const provider = this.getProviders(document).map(p => p.provider).find(p => typeof p.resolveCodeAction === 'function')
+    if (provider === undefined) {
+      return action
+    }
+    const resolved: CodeAction = await provider.resolveCodeAction(action, token)
+    if (resolved.clientId !== action.clientId) {
```
fannheyward, line 35: Will the `resolved` be null, need to check before access clientId.

fannheyward (changes requested):

src/provider/codeActionmanager.ts
adamansky: Sure https://github.com/Softmotions/coc.nvim/commit/05897a8f741d6060beba4481a4ae3cf43e604c17

adamansky: Also recursive call is removed https://github.com/Softmotions/coc.nvim/commit/5547b0d5509dcffa4c424fb0bf520176774b3d9d

PR #3158, opened 2021-06-23
fix(languages): codeActionmanager

fannheyward:


fannheyward:
@chemzqm the `test (ubuntu-latest, 10)` action is already removed, should be changed to 12 or 14.

PR #3186, opened 2021-07-01
feat(semanticTokens): provide basic default colors

sam-mccall:
Not sure if this was omitted intentionally, but the fact that there was no highlighting at all by default threw me for a loop!

---

This maps each of the "standard"[1] CocSem_* highlight groups to the
semantically-closest standard vim group, e.g. CocSem_enum => Type.

This means that a user with coc.nvim and a colorscheme enabled will get
some basic highlighting out of semanticTokens. Not enough to distinguish
everything, but at least e.g. types vs variables.
Today you get nothing by default, as the CocSem_* groups are undefined.

The possible downside: AFAICS coc.nvim doesn't apply highlight groups that
don't exist, so defining these groups means we'll pay the performance
cost of applying the styles even if the benefit is small.

[1] Specifically these are the SemanticTokenTypes defined in LSP:
https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#semanticTokenTypes
(though language servers can support others)

### DIFF
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -2733,6 +2733,38 @@ CocListPath						*CocListPath*
 
   Used for current cwd in statusline of CocList.
 
+CocSem_*						*CocSem*
+
+  Used for words whose semantics are provided by the language server, which
+  typically analyzes source code more deeply than vim's builtin syntax parser.
+
+  Basic defaults are provided for the "standard" token kinds, but any kind
+  supported by the server can be highlighted by creating the highlight group.
+
+  Default (disable by setting `g:coc_default_semantic_highlight_groups = 0`):
+  `hi default link CocSem_namespace Identifier`
+  `hi default link CocSem_type Type`
+  `hi default link CocSem_class Structure`
+  `hi default link CocSem_enum Type`
+  `hi default link CocSem_interface Type`
+  `hi default link CocSem_struct Structure`
+  `hi default link CocSem_typeParameter Type`
+  `hi default link CocSem_parameter Identifier`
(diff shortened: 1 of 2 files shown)


fannheyward:
Yes, semanticTokens feature, include fetch and apply, is added in 3.16 PR, but the highlight group is not defined, the main reason is that I don't have a good master on vimL, thank you for your PR, cc @chemzqm.

chemzqm (changes requested):
Need a vim global variable to toggle the default highlights, and doc for semantic highlights.

oblitum:
I think it should be stated whether the toggle should be default ON or OFF.

sam-mccall:
Added some docs for the highlight groups.

From my POV there's not much point providing defaults if they're not actually on by default, the target here is users who haven't/won't gone looking for things to customize. The defaults here are very basic but safe.

PTAL!

(Honestly having the toggle seems pretty marginal to me: we already have an option to turn semantic highlighting off, so this is only relevant to people who want highlighting *on*, want some/all standard highlight kinds *not* to be highlighted, and don't want to list those in their config in order to unbind/rebind them)

oblitum:
@sam-mccall https://github.com/neoclide/coc.nvim/issues/3217#issuecomment-880629631

Regarding the toggle, thanks for adding, it's rather necessary so people don't need to look up the specific option on each language server. Currently I could not even disable it with current coc-clangd's `clangd.semanticHighlighting`, it's ineffective, and the flag avoids having to list these groups in my init to have the feature disabled, which currently is [presenting some issues](https://github.com/neoclide/coc.nvim/issues/3217).

PR #3187, opened 2021-07-02
feat(typings): more typings

fannheyward:
CallHierarchyMiddleware
SemanticTokensMiddleware
LinkedEditingRangeMiddleware

RegistrationType
RequestType

### DIFF
--- a/src/language-client/client.ts
+++ b/src/language-client/client.ts
@@ -4339,4 +4339,13 @@ export abstract class BaseLanguageClient {
     }
     this.error(`Request ${type.method} failed.`, error)
   }
+
+  public logFailedRequest(type: any, error: any): void {
+    // If we get a request cancel don't log anything.
+    if (error instanceof ResponseError && error.code === LSPErrorCodes.RequestCancelled) {
+      // TODO: handle LSPErrorCodes.ContentModified
+      return
+    }
+    this.error(`Request ${type.method} failed.`, error)
+  }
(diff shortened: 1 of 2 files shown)


src/language-client/client.ts
```diff
@@ -4339,13 +4339,4 @@ export abstract class BaseLanguageClient {
     }
     this.error(`Request ${type.method} failed.`, error)
   }
-
-  public logFailedRequest(type: any, error: any): void {
```
chemzqm, line 4343: Don't remove this method, old language client may throw error.

chemzqm (changes requested):

typings/index.d.ts
```diff
@@ -7689,6 +7852,14 @@ declare module 'coc.nvim' {
      */
     workDoneToken?: ProgressToken
   }
+  export class RegistrationType<RO> {
```
chemzqm, line 7855: Don't use export since coc.nvim doesn't export this class.

typings/index.d.ts
```diff
@@ -7574,65 +7719,83 @@ declare module 'coc.nvim' {
      */
     dispose(): void
   }
+
+  export class ParameterStructures {
```
chemzqm, line 7723: Don't use export since coc.nvim doesn't export this class.

chemzqm (changes requested):

PR #3188, opened 2021-07-02
fix(diagnostic): highlight hints, deprecated and unused text with their own unique hlgroups

dmitmel:
This PR fixes the remaining entries in `DiagnosticHighlight` to use unique hlgroups, namely `Coc<Something>Highlight`, instead of reusing existing ones. Note that I say "remaining" because that enum was using `CocErrorHighlight` and `CocWarningHighlight` from the beginning (commit df7a1d94233bac9e26d593fe11f4f14b049e0a14, [relevant code](https://github.com/neoclide/coc.nvim/commit/df7a1d94233bac9e26d593fe11f4f14b049e0a14#diff-23ccea505c03101016c60ed7cbc58543c16d244e9b2fd5c3446ca3f7f7ced9bbR17-R24)), then `CocInfoFloat` was correctly replaced with `CocInfoHighlight` in commit 4cd2b40390a2dadb56baf3135064f74b148c9211, and finally I replace the remaining hlgroup for the "hint" severity with `CocHintHighlight` in this PR.

Additionally, I added the `CocDeprecatedHighlight` and `CocUnusedHighlight` groups for deprecated and unused code respectively, which link to `CocStrikeThrough` and `CocFadeOut`. This was done both for the sake of consistency, and future-proofing in case the `CocStrikeThrough` and `CocFadeOut` groups may come in handy for text formatting in other unrelated things, similarly to hlgroups such as `CocBold`, `CocItalic`, `CocUnderline` and others. Because the two new hlgroups are by default just linked, this will have no visible effect on user configurations.

As for `CocHintHighlight`, this PR indeed fixes highlighting of hints, which right now look like:

image

And here is how they will look after this PR (which was the behavior before commit df7a1d94233bac9e26d593fe11f4f14b049e0a14):

image

As you can see, their highlighting of the relevant source code is now consistent with errors because by default all `Coc*Highlight` groups are linked to `CocUnderline`.

### DIFF
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -2510,7 +2510,7 @@ CocFadeOut						*CocFadeOut*
 
   Default: `hi default CocFadeOut       guifg=#928374 ctermfg=245`
 
-  Used for highlight unnecessary code.
+  Used for faded out text, such as for highlighting unnecessary code.
 
 CocStrikeThrough 					*CocStrikeThrough*
 
@@ -2621,6 +2621,18 @@ CocHintHighlight					*CocHintHighlight*
   Default: `hi default link CocHintHighlight   CocUnderline`
 
   Used for hint text.
+
+CocDeprecatedHighlight					*CocDeprecatedHighlight*
+
+  Default: `hi default link CocDeprecatedHighlight   CocStrikeThrough`
+
+  Used for usages of deprecated API.
+
+CocUnusedHighlight					*CocUnusedHighlight*
+
+  Default: `hi default link CocUnusedHighlight   CocFadeOut`
+
+  Used for unnecessary code.
 
 CocHighlightText					*CocHighlightText*
 
--- a/plugin/coc.vim
+++ b/plugin/coc.vim
@@ -369,6 +369,8 @@ function! s:Hi() abort
   hi default link CocWarningHighlight CocUnderline
   hi default link CocInfoHighlight    CocUnderline
   hi default link CocHintHighlight    CocUnderline
+  hi default link CocDeprecatedHighlight CocStrikeThrough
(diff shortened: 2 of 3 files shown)


oblitum:
@fannheyward do you think it's a good idea to associate this to [SemanticTokenModifiers.deprecated](https://github.com/neoclide/coc.nvim/blob/7089ffb068ecb1059d15b3766b24280b74dac970/src/language-client/semanticTokens.ts#L92)? I've noticed deprecated strikethrough API's through diagnostics, it's quite useful, but some language servers are only providing it at semantic token side solely, not diagnostic. `gopls` seems to support this both through semantic token modifier and completion item tag:

- https://github.com/golang/tools/commit/edbe9bef0442891a979bf6a16ad7c4c13dd57a39

`rust-analyzer` also seem to support this through semantic token, but sadly, I tried both `coc-rust-analyzer` _and_ `coc-go` to no avail. I've checked their semantic tokens response in the verbose output and the deprecated modifier was never present, I don't understand why, not sure that's a bug in coc.nvim's semantic tokens.

Anyway, back to the topic, I suspect a reasonable deprecated semantic token modifier implementation would be (given https://github.com/neoclide/coc.nvim/tree/feat/use-ts-hi-group), that it picked `CocDeprecatedHighlight` in case `TSDeprecated` is empty, otherwise pick `TSDeprecated`.

fannheyward:
> that's a bug in coc.nvim's semantic tokens

@oblitum  Yes, coc.nvim's semanticTokens didn't pick tokenModifiers by now. Can you give a go/rust example code? I'll take it this week.

fannheyward:
@oblitum `semanticTokens` has been handled in https://github.com/neoclide/coc.nvim/pull/3355, `TSDeprecated` will be highlighted.

PR #3204, opened 2021-07-09
fix(languages): disable onDidSemanticTokensRefresh

fannheyward:
NOTE: this is just a temporary solution, doesn't really fix the issue.


chemzqm:
It's not fix.

Issue #3206, opened 2021-07-09
triggerCharacter not in options when completion is invoked manually


eliasnorrby:
## Result from CocInfo

```
vim version: NVIM v0.4.4
node version: v16.3.0
coc.nvim version: 0.0.80-2d813b6233
term: tmux
platform: linux
```

## Describe the bug

`triggerCharacter` is not part of the options object when completion is invoked manually.

When `Completion.startCompletion` is called via `Completion.triggerCompletion` (as it will be when `"suggest.autoTrigger": "always"`), the options object is enriched with the `triggerCharacter`:

```typescript
if (pre.length) {
  option.triggerCharacter = pre.slice(-1)
}
logger.debug('trigger completion with', option)
await this.startCompletion(option)
```
[context](https://github.com/neoclide/coc.nvim/blob/2d813b6233f7cb313e5c9d54ae2834348d982366/src/completion/index.ts#L409-L413)

But when completion is invoked manually (i.e. by setting up a binding for `coc#refresh()`), `Completion.startCompletion` is called with the raw options object:

```typescript
this.addAction('startCompletion', option => completion.startCompletion(option))
```
[context](https://github.com/neoclide/coc.nvim/blob/2d813b6233f7cb313e5c9d54ae2834348d982366/src/plugin.ts#L177)

I consider this a bug because `triggerCharacter` is not marked as optional in [`CompleteOption`](https://github.com/neoclide/coc.nvim/blob/2d813b6233f7cb313e5c9d54ae2834348d982366/src/types.ts#L424).

## Why is this an issue?

See neoclide/coc-git#185

When an extension (like `coc-git`) depends on a specific prefix (like '#' for completing GitHub issues), its completion will only work when triggered automatically.

## Reproduce the bug

0. (optional) Set `"suggest.autoTrigger": "none"`
1. Install `coc-git`: `:CocInstall coc-git`
2. Set up a binding for `coc#refresh()`

    ```vim
    inoremap <silent><expr> <TAB>
          \ pumvisible() ? "\<C-n>" :
          \ <SID>check_back_space() ? "\<TAB>" :
          \ coc#refresh()
    inoremap <expr><S-TAB> pumvisible() ? "\<C-p>" : "\<C-h>"
    ```

3. Open a commit message editor in a repo with issues (e.g. `git commit --allow-empty`)
4. Type '#' and press TAB

## A possible solution

[...]

PR #3207, opened 2021-07-09
fix(completion): set triggerCharacters in get_complete_option

eliasnorrby:
Set triggerCharacters in coc#util#get_complete_option so that it's available both when completion is triggered automatically and manually.

closes: #3206


chemzqm:
Fixed by a56c2e80

PR #3212, opened 2021-07-12
feat(extension): fix flag isActive too early

iamcco:


iamcco:
Seems like this fix will have another issue.

Issue #3215, opened 2021-07-13
Error detected while processing coc.vim -> -complete used without -nargs


Whitecx:
## Describe the bug
My vim version 8.2.3150 give the following warning with coc.vim installed
`Error detected while processing .vim/plugged/coc.nvim/plugin/coc.vim:
line  484: E1208: -complete used without -nargsPress ENTER or type command to continue`

Looking at the repo for coc.vim, the existing codebase matches what I have locally. Specifically, this line is the offender
```vim 
command! -nargs=0 -complete=custom,coc#list#names CocListCancel   :call coc#rpc#notify('listCancel', [])
```

Looking at the Vimscript docs, I see that the following attribute is included to allow for autocompletion for arguments:
```vim
-complete=custom,{func} custom completion, defined via {func}
```
However, line 484 of the plugin specifies 0 arguments, which makes the line seem unnecessary and is probably causing the warning. Should -complete-custom be removed?

## Reproduce the bug
.vimrc:
```vim  1 set number
  2 set laststatus=2
  3 set noshowmode
  4 set tabstop=4
  5 set hlsearch
  6 set nocompatible              " required
  7 filetype off                  " required
  8 
  9 call plug#begin('~/.vim/plugged')
 10 
 11 Plug 'neoclide/coc.nvim', {'branch': 'release'}
 12 Plug 'preservim/nerdtree'
 13 call plug#end()
 14 
 15 " Enable folding
 16 set foldmethod=indent
 17 set foldlevel=99
 18 " Enable folding with the spacebar
 19 nnoremap <space> za
 20 
 21 let python_highlight_all=1
 22 syntax on
```

Steps: Open any file w/ vim (v. 8.2.3150)

jenhsun:
Just change .vim/plugged/coc.nvim/plugin/coc.vim line 484's **-nargs=0** to **-nargs=?** and keep **-complete=custom,coc#list#names** solve the problem.

Fix from https://github.com/neoclide/coc.nvim/pull/3211#issuecomment-878179098

chemzqm:
Fixed on latest release branch of coc.nvim.

PR #3211, opened 2021-07-12
Fix command with complete but no nargs

perrin4869:
Maybe it's better to target master branch?
Reference: vim/vim#8544
Closes #3209

PR #3216, opened 2021-07-13
removed -complete from Coclistcancel because it takes 0 args

Whitecx:
Per Vimscript docs, -complete is used to allow autocomplete for arguments, but since -nargs is set to 0, an error was produced when running vim. 

(See https://github.com/neoclide/coc.nvim/issues/3215)


Whitecx:
I see the change was already made in master, but not yet reflected in release
(https://github.com/neoclide/coc.nvim/pull/3211#issuecomment-878179098)

PR #3223, opened 2021-07-20
Fix: [coc.nvim]: Error on "rangeSelect"

adamansky:
Fix of undefined `selectionRange` handler attribute which leads to 
```
[coc.nvim]: Error on "rangeSelect": Cannot read property 'selectRange' of undefined
```
when calling `coc-range-select` command.

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -26,6 +26,7 @@ import CallHierarchy from './callHierarchy'
 import SemanticTokensHighlights from './semanticTokensHighlights/index'
 import Signature from './signature'
 import Symbols from './symbols'
+import SelectionRangeHandler from './selectionRange'
 const logger = require('../util/logger')('Handler')
 
 export interface CurrentState {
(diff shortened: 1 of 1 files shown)


fannheyward (approved):

src/handler/index.ts
```diff
@@ -26,6 +26,7 @@ import CallHierarchy from './callHierarchy'
 import SemanticTokensHighlights from './semanticTokensHighlights/index'
 import Signature from './signature'
 import Symbols from './symbols'
+import SelectionRangeHandler from './selectionRange'
```
fannheyward, line 29: This is not needed as SelectionRange is already imported

src/handler/index.ts
```diff
@@ -83,6 +84,7 @@ export default class Handler {
     this.callHierarchy = new CallHierarchy(nvim, this)
     this.documentHighlighter = new Highlights(nvim, this)
     this.semanticHighlighter = new SemanticTokensHighlights(nvim, this)
+    this.selectionRange = new SelectionRangeHandler(nvim, this);
```
fannheyward, line 87: `this.selectionRange = new SelectionRange(nvim, this)`

fannheyward (changes requested):

src/handler/index.ts
adamansky: done

src/handler/index.ts
adamansky: done

fannheyward (approved):

PR #3227, opened 2021-07-21
Allow whitespace as trigger characters

NoahTheDuke:
[coc-clojure](https://github.com/noahtheduke/coc-clojure) uses spaces.

### DIFF
--- a/src/provider/signatureManager.ts
+++ b/src/provider/signatureManager.ts
@@ -7,12 +7,7 @@ import { v4 as uuid } from 'uuid'
 export default class SignatureManager extends Manager<SignatureHelpProvider> implements Disposable {
 
   public register(selector: DocumentSelector, provider: SignatureHelpProvider, triggerCharacters?: string[]): Disposable {
-    let characters = triggerCharacters.reduce((p, c) => {
-      if (/^\s+$/.test(c)) {
-        return p.concat(c)
-      }
-      return p.concat(c.split(/\s*/g).filter(s => s.length > 0))
-    }, [] as string[])
+    let characters = triggerCharacters.reduce((p, c) => p.concat(c.length == 1 ? c : c.split(/\s*/g)), [] as string[])
     let item: ProviderItem<SignatureHelpProvider> = {
       id: uuid(),
       selector,
--- a/src/provider/signatureManager.ts
+++ b/src/provider/signatureManager.ts
@@ -7,7 +7,12 @@ import { v4 as uuid } from 'uuid'
 export default class SignatureManager extends Manager<SignatureHelpProvider> implements Disposable {
 
   public register(selector: DocumentSelector, provider: SignatureHelpProvider, triggerCharacters?: string[]): Disposable {
-    let characters = triggerCharacters.reduce((p, c) => p.concat(c.split(/\s*/g)), [] as string[])
+    let characters = triggerCharacters.reduce((p, c) => {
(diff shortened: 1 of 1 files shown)


src/provider/signatureManager.ts
```diff
@@ -7,7 +7,12 @@ import { v4 as uuid } from 'uuid'
 export default class SignatureManager extends Manager<SignatureHelpProvider> implements Disposable {
 
   public register(selector: DocumentSelector, provider: SignatureHelpProvider, triggerCharacters?: string[]): Disposable {
-    let characters = triggerCharacters.reduce((p, c) => p.concat(c.split(/\s*/g)), [] as string[])
```
chemzqm, line 10: Change this line to
```
let characters = triggerCharacters.reduce((p, c) => p.concat(c.startsWith(' ') ? [' ']: c.split(/\s*/g)), [] as string[])
```

chemzqm (changes requested):

src/provider/signatureManager.ts
NoahTheDuke: The goal was to allow for characters other than <space>, such as <newline>. The spec isn't clear either way. What do you think?

chemzqm: Use

``` ts
let characters = triggerCharacters.reduce((p, c) => p.concat(c.length == 1 ? c : c.split(/\s*/g)), [] as string[])
```

NoahTheDuke: Cool, thank you.

src/provider/signatureManager.ts
```diff
@@ -7,7 +7,7 @@ import { v4 as uuid } from 'uuid'
...
+    let characters = triggerCharacters.reduce((p, c) => p.concat(c.length == 1 ? c : c.split(/\s*/g)), [] as string[])
```
chemzqm, line 10: Should be `p.concat(c.length == 1 ? [c]`

chemzqm (changes requested):

src/provider/signatureManager.ts
NoahTheDuke:
```suggestion
    let characters = triggerCharacters.reduce((p, c) => p.concat(c.length == 1 ? [c] : c.split(/\s*/g)), [] as string[])
```

NoahTheDuke: Oops, missed your edit. Thanks so much.

oblitum:
@NoahTheDuke Hi. I'm just trying out Clojure for the first time (but I'm a long time coc.nvim user). Does your PR still work in latest coc.nvim? I'm using your coc-clojure extension and this code won't trigger signature help (I was unable to get signatureHelp to be triggered, not sure whether I need to write some special Clojure code declaration that would cause signatures to show up):

```clojure
(defn greet  [name]  (str "Hello, " name) )

(greet <space here triggers nothing>
```

oblitum:
@NoahTheDuke ping?

oblitum:
I'm able to get signatureHelp from clojure-lsp only through manual `CocActionAsync('showSignatureHelp')` from insert mode:

gif-2021-08-28-112332

Automatically after whitespace nothing comes up:
gif-2021-08-28-112540

oblitum:
So, checking this again, clojure-lsp actually doesn't use whitespace as trigger characters for signature help.

PR #3243, opened 2021-07-26
feat(handler): definition content on hover

fannheyward:
截屏2021-07-28 下午4 26 57

VSCode: 

截屏2021-07-26 下午5 18 38

cc @chemzqm , some Todos:

~1. I put `definitionHover` in `handler` directly because we can use `locations` and `hover` to fetch and preview content, do we need to put into a new separate handler?~
~2. how to handle multiple definition locations? VSCode shows a `Click to show {0} definitions.` action, but we can't put action in hover by now~
~3. the definition content needs to be marked to markdown for highlighting~

### DIFF
--- a/src/handler/index.ts
+++ b/src/handler/index.ts
@@ -1,5 +1,5 @@
 import { NeovimClient as Neovim } from '@chemzqm/neovim'
-import { CancellationToken, CancellationTokenSource, Disposable, Position } from 'vscode-languageserver-protocol'
+import { CancellationToken, CancellationTokenSource, Disposable, Hover, MarkupKind, Position } from 'vscode-languageserver-protocol'
 import { TextDocument } from 'vscode-languageserver-textdocument'
 import events from '../events'
 import languages from '../languages'
@@ -16,7 +16,7 @@ import Commands from './commands'
 import Fold from './fold'
 import Format from './format'
 import Highlights from './highlights'
-import HoverHandler from './hover'
+import HoverHandler, { HoverTarget } from './hover'
 import Links from './links'
 import Locations from './locations'
 import Refactor from './refactor/index'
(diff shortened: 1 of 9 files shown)


src/handler/index.ts
```diff
@@ -116,6 +116,40 @@ export default class Handler {
...
+      const endLine = end.line - start.line >= 8 ? start.line + 8 : end.line
+      const lines = doc.getLines(start.line, endLine)
+      if (lines.length) {
+        const defHover: Hover = {
+          range: loc.range,
+          contents: { kind: MarkupKind.PlainText, value: lines.join('\n') }
+        }
+        hovers.splice(0, 0, defHover)
+      }
+      await this.hover.previewHover(hovers, target)
```
chemzqm, line 149: Move the code block to `handler/hover.ts` and refactor `previewHover` to support `hovers: (Hover | Documentation)[]`

src/handler/index.ts
```diff
@@ -116,6 +116,40 @@ export default class Handler {
...
   }
 
+  public async definitionHover(target: HoverTarget): Promise<void> {
+    const { doc, position } = await this.getCurrentState()
+    this.checkProvier('hover', doc.textDocument)
+    await doc.synchronize()
+    const tokenSource = new CancellationTokenSource()
+    const hovers =  await languages.getHover(doc.textDocument, position, tokenSource.token)
+
+    const locs = await this.locations.definitions()
```
chemzqm, line 126: This can't be used since we have to use `DefinitionLink` here,   you need refactor `DefinitionManager` module to export a method for get definition links, and make add new method to `languages` for provide definition links.

chemzqm (changes requested):

chemzqm:
> how to handle multiple definition locations? VSCode shows a Click to show {0} definitions. action, but we can't put action in hover by now

Show all of them should be fine.

src/handler/index.ts
fannheyward: I'm a little confused, current `DefinitionManager` provides definitions as `Location`, you mean we should only use `DefinitionLink` for definition on hover? How about that LS doesn't return DefinitionLink?

fannheyward: Will ignore if no definition links found.

chemzqm:
I've moved the code block to the end of buffer since syntax item with transparent could break syntax highlighting and I can't figure out how to fix it.

PR #3274, opened 2021-08-07
fix(diagnostic): keep updating diagnostic at least one time

kevinhwang91:
For now, updating diagnostic may be canceled which make the diagnostic
information is missing. 

Below video show what the issue is.

https://user-images.githubusercontent.com/17562139/128592201-c1b27c18-112a-4de4-aae5-794d36c120ee.mp4

Remove the onChange method to make sure that the changed buffer will
update diagnostic at least one time


chemzqm:
Your language server should send new diagnostics after document change.

PR #3281, opened 2021-08-10
[RFC] Support callHierarchy in tree view

### LINKED ISSUE #1649: Support call hierarchies

No problem other than tracking feature parity with tsserver (here is the companion issue for [coc-tsserver](https://github.com/neoclide/coc-tsserver/issues/128)) and VSCode.

Support for retrieving call hierarchies in tree form for named callables. 

```
class MainThing {
  static do() {
    utilA();
  }
}

//    X-- cursor 
const doLotsOfThings () => {
  MainThing.do();
  utilB();
}

// 
// :CocList callHierarchy --maxDepth 3

OUTPUT

--------------------------------------------------------------
doLotsOfThings()
  calls
    MainThing.do()
      utilA()
    utilB()
  is called by
    Scheduler.run

```

I believe the minimal required UI would be list-based, similar to the 'outline' list. It could take a "max depth" argument to control the indentation.

Actions on list items might include: 
- go to definition
- get references
- re-request hierarchy using same args but with this callable as the root.

I _think_ you could do this with existing LSP features in a coc extension, but without language-server-level support there would likely be a problem with N+1 queries.

**Additional context**

I'm not sure which API the project's maintainers prioritize, that of VSCode, Microsoft's LSP spec, or both. So I'm likewise unsure how important it is that as of this writing, the [proposal issue](https://github.com/microsoft/language-server-protocol/issues/468) has yet to be adopted by Microsoft, though the latest release of VSCode supports it.

chemzqm:
Could works now, closes #1649

Added actions:

- `showIncomingCalls`
- `showOutgoingCalls`

Added commands:

- `document.showIncomingCalls`
- `document.showOutgoingCalls`

Todos:

- [x] Actions of tree item.
- [x] Show `containerName` after label like VSCode.
- [x] Highlight range after location jump.
- [ ] Tests

### DIFF
--- a/src/handler/callHierarchy.ts
+++ b/src/handler/callHierarchy.ts
@@ -1,7 +1,24 @@
 import { Neovim } from '@chemzqm/neovim'
-import { CallHierarchyIncomingCall, Range, CallHierarchyItem, CallHierarchyOutgoingCall, CancellationTokenSource } from 'vscode-languageserver-protocol'
+import { CallHierarchyIncomingCall, Position, Range, CallHierarchyItem, CallHierarchyOutgoingCall, CancellationTokenSource, SymbolKind, SymbolTag, Location, Disposable, CancellationToken } from 'vscode-languageserver-protocol'
 import languages from '../languages'
-import { HandlerDelegate } from '../types'
+import workspace from '../workspace'
+import { ConfigurationChangeEvent, HandlerDelegate } from '../types'
+import BasicTreeView from '../tree/TreeView'
+import { TreeDataProvider, TreeItem, TreeItemCollapsibleState, TreeItemIcon } from '../tree/index'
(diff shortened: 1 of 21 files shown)


fannheyward:
截屏2021-08-11 上午10 29 28

I can close my local branch and work on this now :D

src/handler/callHierarchy.ts
```diff
@@ -127,6 +144,28 @@ export default class CallHierarchyHandler {
         element.children = items
         return items
       },
+      resolveActions: () => {
```
fannheyward, line 154: Do we need to add the `Dismiss` action as VSCode?

截屏2021-08-18 下午2 24 22

chemzqm: I don't think it's necessary.

PR #3294, opened 2021-08-18
feat: add coc_seamtic_highlight_filetypes

fannheyward:


chemzqm:
It doesn't work, the highlight commands runs on coc.nvim load.

PR #3297, opened 2021-08-19
fix(hover): shift-left definition content

fannheyward:
The lines from DefinitionLink maybe indented, I just check the first line whether it starts with space/tab, and shift left all lines if found.

Before:

截屏2021-08-19 下午3 29 23

After:

截屏2021-08-19 下午3 28 30

### DIFF
--- a/src/handler/hover.ts
+++ b/src/handler/hover.ts
@@ -114,7 +114,10 @@ export default class HoverHandler {
         if (!def.targetRange) continue
         const { start, end } = def.targetRange
         const endLine = end.line - start.line >= 100 ? start.line + 100 : (end.character == 0 ? end.line - 1 : end.line)
-        const lines = await readLines(def.targetUri, start.line, endLine)
+        let lines = await readLines(def.targetUri, start.line, endLine)
+        if (lines[0].search(/\S/) > 0) {
+          lines = lines.map(l => l.substring(lines[0].search(/\S/)))
+        }
         hovers.push({ content: lines.join('\n'), filetype: doc.filetype })
       }
     }
--- a/src/handler/hover.ts
+++ b/src/handler/hover.ts
@@ -115,9 +115,9 @@ export default class HoverHandler {
         const { start, end } = def.targetRange
         const endLine = end.line - start.line >= 100 ? start.line + 100 : (end.character == 0 ? end.line - 1 : end.line)
         let lines = await readLines(def.targetUri, start.line, endLine)
-        const indent = lines[0].search(/\S/)
+        const indent = lines[0].length - lines[0].trimStart().length
         if (indent) {
-          lines = lines.map(l => l.search(/\S/) > 0 ? l.substring(indent) : l)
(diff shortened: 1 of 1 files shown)


src/handler/hover.ts
```diff
@@ -114,7 +114,10 @@ export default class HoverHandler {
         if (!def.targetRange) continue
         const { start, end } = def.targetRange
         const endLine = end.line - start.line >= 100 ? start.line + 100 : (end.character == 0 ? end.line - 1 : end.line)
-        const lines = await readLines(def.targetUri, start.line, endLine)
+        let lines = await readLines(def.targetUri, start.line, endLine)
+        if (lines[0].search(/\S/) > 0) {
+          lines = lines.map(l => l.substring(lines[0].search(/\S/)))
```
chemzqm, line 119: It's possible to remove code that not indent and you should not call `lines[0].search` too many times.

chemzqm (changes requested):

src/handler/hover.ts
```diff
@@ -114,7 +114,11 @@ export default class HoverHandler {
         if (!def.targetRange) continue
         const { start, end } = def.targetRange
         const endLine = end.line - start.line >= 100 ? start.line + 100 : (end.character == 0 ? end.line - 1 : end.line)
-        const lines = await readLines(def.targetUri, start.line, endLine)
+        let lines = await readLines(def.targetUri, start.line, endLine)
+        const indent = lines[0].search(/\S/)
```
chemzqm, line 118: Bad regex, should check start position only

src/handler/hover.ts
```diff
@@ -114,7 +114,11 @@ export default class HoverHandler {
...
+        if (indent) {
+          lines = lines.map(l => l.search(/\S/) > 0 ? l.substring(indent) : l)
```
chemzqm, line 120: Use `line.startsWith` instead

chemzqm (changes requested):

src/handler/hover.ts
fannheyward: Yes, we can use `line.startsWith` to check whether line indented with space/tab,  but I need the non-space position in first, to cut other lines. That's why I use `/\S/` to get the position. Do I miss-understand something you said?

chemzqm: What if first line have four space indent but the line after have two space?

chemzqm: And the `search()` method could returns `-1` which you should ignore.
My suggestion is use `match` get indent text and check the line with `startsWith`

chemzqm (changes requested):

src/handler/hover.ts
fannheyward: > use match get indent text

Got it.

fannheyward: @chemzqm  I found it's hard to use `match` to get the first line indent, so I use `line.length - line.trimStart().length` to get the indent.

chemzqm:
It still possible trim none space text of line, I'll make a fix.

PR #3302, opened 2021-08-21
fix(float.vim): prompt canceled in neovim will move cursor

kevinhwang91:
When using `<esc>` to cancel the prompt window and go back to the previous window,
cursor shift one left.

Look like it's a bug for closing floating window in insert mode.


chemzqm:
The issue happens on normal mode as well, I will make a fix.

kevinhwang91:
After my patch, press <C-c> and then <Esc> or <Esc> directly, the cursor keep the previous position.
I don't understand what you say happening on normal mode.

PR #3317, opened 2021-08-27
fix(completion): restart completion to handle indent changed

kevinhwang91:
Coc shouldn't stop completion directly if the indent changed, restart completion is better.


chemzqm:
Test needed.

kevinhwang91:
OK, it may be slow for me to add a test, but I will give it a try.

kevinhwang91:
I found the other issues. I will open this PR if I think everything is fine.

chemzqm:
Need to make sure `this.config.autoTrigger` is `always` as well.

PR #3325, opened 2021-08-31
Add option to show border on hover window

yuki-yano:
Added an option to set the border for the hover window.
I think floating windows with borders are better for viewing information.

The default is false, so I don't think it will affect existing users.

CleanShot 2021-08-31 at 11 54 16

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -766,7 +766,7 @@
       "default": true,
       "description": "Automatically hide hover float window on CursorMove or InsertEnter."
     },
-    "hover.displayBorder": {
+    "hover.floatBorder": {
       "type": "boolean",
       "default": false,
       "description": "Show border on hover float window."
--- a/data/schema.json
+++ b/data/schema.json
@@ -771,6 +771,11 @@
       "default": false,
       "description": "Show border on hover float window."
     },
+    "hover.floatHighlight": {
+      "type": "string",
+      "default": "CocFloating",
+      "description": "Highlight group for hover window/popup, default to 'CocFloating'"
+    },
     "dialog.maxHeight": {
       "type": "number",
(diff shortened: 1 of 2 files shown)


fannheyward:
Thank you for the PR, I prefer the configuration name to `hover.floatBorder`.

yuki-yano:
Thanks, I'll try to fix it that way.
Also, I think I'm going to add `hover.floatHighlight`, what do you think?

fannheyward:
> add hover.floatHighlight

Do you mean something like `dialog.floatHighlight`?

yuki-yano:
When I set CocFloating to change the highlight of a hover window, the highlight is changed to the window that does not have a border.

For example, I want to set a different highlight for the info of the completion.
(I'd also like to change it so that I can set the border on the info of the completion)

yuki-yano:
Sorry, I'll try dialog.floatHighlight.

yuki-yano:
Hmmm, I tried but it was not possible to set the highlight of the hover window and other highlights separately.
Is it undesirable to be able to set the highlight of the hover window?

yuki-yano:
If you set the border in the floating window, you will want to set the highlight separately, as it is easy to distinguish even if the color is close to Normal.

This may not need to be included in this change.
Would it be better to create individual PRs?

yuki-yano:
As a similar feature, VSCode has such a setting item.

```
"workbench.colorCustomizations": {
    "editorHoverWidget.border": "#ff0000",
    "editorHoverWidget.background": "#ff0000"
    "editorHoverWidget.foreground": "#ff0000"
}
```

fannheyward (approved):

chemzqm:

PR #3326, opened 2021-08-31
Add option to show border on suggest window/popup window

yuki-yano:
Added an option to show borders in the suggest window, similar to #3325.

CleanShot 2021-08-31 at 16 42 00


fannheyward (approved):

chemzqm:

PR #3335, opened 2021-09-01
Add option to show border on diagnostic float/popup window

yuki-yano:
Added an option to show borders in the suggest window, similar to #3325 and #3326.

CleanShot 2021-09-02 at 08 48 56


fannheyward (approved):

chemzqm:

PR #3336, opened 2021-09-02
fix(configuration): create config file if not exist

fannheyward:
If `coc-settings.json` file is not exists, `configuration.update()` will raise exception. This fix will create the file from `readFileSync` if not exists, the default flag is `r`, changed to `a+`:

- `r`: Open file for reading. An exception occurs if the file does not exist.
- `a+`: Open file for reading and appending. The file is created if it does not exist.

https://nodejs.org/api/fs.html#fs_file_system_flags

closes https://github.com/clangd/coc-clangd/issues/344

### DIFF
--- a/src/configuration/shape.ts
+++ b/src/configuration/shape.ts
@@ -22,7 +22,7 @@ export default class ConfigurationProxy implements ConfigurationShape {
     let file = workspace.getConfigFile(target)
     if (!file) return
     let formattingOptions: FormattingOptions = { tabSize: 2, insertSpaces: true }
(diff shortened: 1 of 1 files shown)


chemzqm:
The parent folder of user config file may not exists as well, my suggestion is prompt user to create.

fannheyward:
> The parent folder of user config file may not exists as well, my suggestion is prompt user to create.

Sorry for the delay, I missed the comment. The parent folder is `~/.config/nvim` or `~/.vim`, they couldn't be not exists.

PR #3346, opened 2021-09-04
feat: add SnippetString

weirongxu:


typings/index.d.ts
```diff
@@ -7498,6 +7498,83 @@ declare module 'coc.nvim' {
...
+    );
+
+    /**
+     * Builder-function that appends the given string to
+     * the {@link SnippetString.value `value`} of this snippet string.
+     *
+     * @param string A value to append 'as given'. The string will be escaped.
+     * @return This snippet string.
+     */
+    appendText(string: string): SnippetString;
```
fannheyward, line 7532: Please remove `;` from the end, same as others in typings.

fannheyward (changes requested):

PR #3404, opened 2021-09-28
fix(snippets): shift range to left as far as possible

kevinhwang91:
coc can't obtain the changing byte event for buffer and use its
`getChange` function as a workaround. The function may get the wrong
range at some edge cases so that it makes `adjustTextEdit` work
unexpected.

Let me explain what issue I'm talking about. The column is 1-index at
the below assumption:

1. The raw text is `foo0bar1`, and then type `.` to trigger the completion menu, **the postfix candidate with an additional TextEdit.del** is showed and preselected;
2. After selecting the postfix candidate, the text line becomes `foo0bar1.foo = foo0bar1`;
3. After applying additional edit which deletes text range from column 1 to 9 meaning `foo0bar1.` from the head, the text line become `foo = foo0bar1` that is the final text;
4. But coc generate the changed range from column 4 to 12, the changed text is `0bar1.foo` instead of `foo0bar1.`;
5. `adjustTextEdit` function misunderstand the range;

### DIFF
--- a/src/__tests__/modules/completion.test.ts
+++ b/src/__tests__/modules/completion.test.ts
@@ -482,6 +482,34 @@ describe('completion TextChangedP', () => {
     expect(col).toBe(12)
   })
 
+  it('should fix cursor position and keep placeholder with snippet on additionalTextEdits', async () => {
+    let doc = await helper.createDocument()
+    let text = 'foo0bar1'
+    await nvim.setLine(text)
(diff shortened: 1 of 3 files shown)


fannheyward (approved):

chemzqm:
Seems we should fix `changes` created while apply edits https://github.com/neoclide/coc.nvim/blob/master/src/model/document.ts#L318

Issue #757, opened 2019-05-05
Wrong languageId passed for TeX and friends


astoff:
Coc sends the wrong `languageId` parameter for LaTeX and plain TeX files. According to the LSP spec, the values should be `latex` for LaTeX and `tex` for plain, but coc (and vim) refers to these types as `tex` and `plaintex` respectively.

To reproduce, one can use the [server configuration](https://github.com/neoclide/coc.nvim/wiki/Language-servers#latex) that I recently added to the wiki.

chemzqm:
You have to use `g:coc_filetype_map` for now.

chemzqm:
Added to wiki page.

valentjn:
Does “for now” mean this issue is going to be fixed eventually?

`tex` and `plaintex` are [Vim filetypes](https://neovim.io/doc/user/filetype.html), while `latex` and `tex` are [LSP language identifiers](https://microsoft.github.io/language-server-protocol/specification#textDocumentItem). Mixing those two (especially using `tex` as LSP language identifier for L<sup>A</sup>T<sub>E</sub>X files) is against the recommendations of the LSP spec. It shouldn't be the task of the user to fix this manually, as the filetype (`tex`, which means L<sup>A</sup>T<sub>E</sub>X) is correctly determined by Vim for L<sup>A</sup>T<sub>E</sub>X files.

oblitum:
I agree. I'm for reopening/fixing.

astoff:
FWIW, I think you should fix the `latex` case, but I would suggest keeping `plaintex` instead of `tex` as language id. It's not clear what the language id table linked above means by TeX.

Also, Digestif (which AFAIK is the only server that supports plain TeX) already understands `plaintex` and maps `tex -> latex` (because of vim), so you would break it you you mapped `plaintex -> tex`.

valentjn:
@astoff I'll open a PR for using the language ID `latex` instead of the language ID `tex`. This would only break servers that somehow don't support `latex`, but that still support L<sup>A</sup>T<sub>E</sub>X documents via the language ID `tex`. This should be very few servers, and those could be easily fixed by adding support for the language ID `latex`, which is the LSP recommendation, anyway.

PR #3433, opened 2021-10-17
feat(document): use `latex` language ID instead of `tex` (#757)

valentjn:
Use `latex` as LSP language ID for L<sup>A</sup>T<sub>E</sub>X documents (instead of `tex`) as recommended by the [LSP specification](https://microsoft.github.io/language-server-protocol/specification#textDocumentItem). This allows using LSP servers that act on L<sup>A</sup>T<sub>E</sub>X documents, but not on plain T<sub>E</sub>X documents. The LSP language ID/Vim filetype `plaintex` for plain T<sub>E</sub>X documents is left unchanged.

I'm not quite sure if I did it right in `list/basic.ts`. Having a function `getFiletype(filetype: string): string` is pretty confusing (why would I get a filetype, if I have to pass it as a argument...). The code would benefit from naming variables for Vim filetypes and variables for LSP language IDs differently (not both `filetype`), but that's outside the scope of the issue/of this PR.

Linted and tested according to the contribution guidelines (although a lot of tests currently fail on `master`), plus manually E2E-tested with my language server [LT<sub>E</sub>X LS](https://github.com/valentjn/ltex-ls).

See #757

### DIFF
--- a/src/__tests__/list/basicList.test.ts
+++ b/src/__tests__/list/basicList.test.ts
@@ -52,6 +52,7 @@ describe('getFiletype()', () => {
   it('should get filetype', async () => {
     expect(getFiletype('javascriptreact')).toBe('javascript')
     expect(getFiletype('typescriptreact')).toBe('typescript')
+    expect(getFiletype('latex')).toBe('tex')
     expect(getFiletype('foo.bar')).toBe('foo')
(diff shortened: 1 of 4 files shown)


oblitum:
@valentjn there's now a new release, so you may update vscode-ltex docs on nvim/nvim support through coc.nvim.

PR #3456, opened 2021-11-07
Add diagnostic.messageTarget none

Gee19:
Adds a third option `none` to `diagnostic.messageTarget`

This makes browsing files with diagnostic errors/warnings less noisy or jarring, especially when there are multiple diagnostics on a single line.

Makes for a much better experience especially when combined with the following:

vimrc:
```
set mouse=a
```

coc-settings.json:
```
"diagnostic.virtualText": true,
"diagnostic.virtualTextCurrentLineOnly": false,
"diagnostic.checkCurrentLine": true,
```

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -611,7 +611,7 @@
       "type": "string",
       "description": "Diagnostic message target.",
       "default": "float",
-      "enum": ["echo", "float"]
+      "enum": ["echo", "float", "none"]
     },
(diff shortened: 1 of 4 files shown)


fannheyward:
@Gee19 hi, I didn't get your point on `messageTarget` to `none`. What's you want is: you only want to display diagnostic with virtualText?

Issue #3473, opened 2021-11-20
Popup text colour matches background


ayroblu:
## Result from CocInfo

```
## versions

vim version: VIM - Vi IMproved 8.2 8023550
node version: v17.0.1
coc.nvim version: 0.0.80-03c9add7cd
coc.nvim directory: /Users/blu/.vim/plugged/coc.nvim
term: tmux
platform: darwin

## Log of coc.nvim

2021-11-20T16:04:09.397 INFO (pid:74127) [plugin] - coc.nvim 0.0.80-03c9add7cd initialized with node: v17.0.1 after 435ms
2021-11-20T16:04:09.446 INFO (pid:74127) [services] - registered service "tsserver"
2021-11-20T16:04:11.111 INFO (pid:74127) [attach] - receive notification: showInfo []
```

## Describe the bug

Popup text colour matches background as per this image:

![image](https://user-images.githubusercontent.com/4915682/142733069-c1a3b45b-afe2-4da8-8277-495df804a7cc.png)

The text is there when I use tmux to select it: 

![image](https://user-images.githubusercontent.com/4915682/142733085-20b0144f-2012-4300-a2cf-5c1d65cfb443.png)

Documentation does show up though, so I'm not sure if it's a tsserver or something else issue

![image](https://user-images.githubusercontent.com/4915682/142733957-4dd55e90-74c1-4415-a531-5f49cf9f79c9.png)

## ~~Reproduce the bug~~ Asking for help

I'm not exactly sure how to debug this issue, but would love some hints, my vimrc is at:
https://github.com/ayroblu/dotfiles/blob/master/.vimrc#L679
with the coc settings specifically at:
https://github.com/ayroblu/dotfiles/blob/master/.vimrc-plugins#L495

I'm using alacritty with the solarised theme as described here: 
https://github.com/ayroblu/dotfiles/blob/master/.alacritty.yml

I do have the same issue on the normal terminal fwiw (different colours, but the text is still unreadable)

oblitum:
This is difficult to ever happen, the colorscheme should be broken somehow regarding floating highlighting, check `:h CocFloating` for learning the highlighting groups you have to tweak.

ayroblu:
I tried changing Pmenu based on the help, but it wasn't working for me
![image](https://user-images.githubusercontent.com/4915682/142741735-b26c1060-b2e0-464c-befe-c49e40d63093.png)
[...]

PR #3477, opened 2021-11-21
fix(highlight): fix regression in highlight colour when only one colour has a "reverse"

ayroblu:
Fix: #3473
This regression was introduced in 5f423c4a81698aac3c83adea7c1adad1549cb853
It affects users of https://github.com/altercation/vim-colors-solarized
Minimal reproduction can be found in the issue, in this comment: https://github.com/neoclide/coc.nvim/issues/3473#issuecomment-974731723

The issue is that there is only one reverse (in CocFloating -> Pmenu), and not in CocInfoSign as per this comment: https://github.com/neoclide/coc.nvim/issues/3473#issuecomment-974746261

## Result

|               | CocInfoSign cterm | CocErrorSign cterm | CocInfoSign gui | CocErrorSign gui |
| -- | -- | -- | -- | -- |
| Before regression | image | image | image | image |
| After regression | image | image | image | image |
| After fix | image | image | image | image |

### DIFF
--- a/autoload/coc/highlight.vim
+++ b/autoload/coc/highlight.vim
@@ -331,10 +331,12 @@ function! coc#highlight#compose_hlgroup(fgGroup, bgGroup) abort
   endif
   let fgId = synIDtrans(hlID(a:fgGroup))
   let bgId = synIDtrans(hlID(a:bgGroup))
-  let guifg = synIDattr(fgId, 'reverse', 'gui') !=# '1' ? synIDattr(fgId, 'fg', 'gui') : synIDattr(fgId, 'bg', 'gui')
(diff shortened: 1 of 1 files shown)


fannheyward (approved):

oblitum:
fwiw, I think, this may still come back in some other specific colorscheme. At the moment code is flipping coins to decide the colors for composing, I think the safest at that would be to calculate color distance and pick colors based on which set gives more contrast (I used to do that kind of color patching a long time ago on my old `.vimrc`, learned it from experience).

fannheyward:
> the safest at that would be to calculate color distance and pick colors based on which set gives more contrast

@oblitum would you like make a PR for this? I'm not very good at the colorscheme.

oblitum:
@fannheyward will try it when I get some time for coding ;-)

PR #3487, opened 2021-11-26
fix(highlight): remove most problematic usage of &winhl

### LINKED ISSUE #1251: Odd highlighting in floating windowns in nvim when you do a Hover
## Result from CocInfo
```
vim version: NVIM v0.4.2
node version: v10.16.3
coc.nvim version: 0.0.74-a84f813105
term: iTerm.app
platform: darwin

## Output channel: tsserver

[Info  - 9:41:48 AM] Forking TSServer
PATH: /<my path stuff>
[Info  - 9:41:48 AM] Started TSServer

{
  "path": "/Users/me/.config/coc/extensions/node_modules/coc-tsserver/node_modules/typescript/lib",
  "_pathLabel": "",
  "_api": {
    "versionString": "3.6.2",
    "version": "3.6.2"
  }
}
```

## Describe the bug

When I issue a `call CocAction('doHover')` the highlighting that is found on the parens around (method) are highlighted in a way that makes it really hard to see depending on your theme. I've noticed this highlighting with nvim's floating window in a way that doesn't really seem to add anything or make sense. Is there a way to turn it off?

## Reproduce the bug

You can reproduce this using the mini.vim that you have listed below.

  ```vim
  set nocompatible
  filetype plugin indent on
  syntax on
  set hidden
  ```

Open up the following js file:
```javascript
const someFunction = (result) => {
  console.info(JSON.parse(result))
}
```

do a call CocAction('doHover')` on parse or any other function and you'll see something that exists like below in the screenshots.

## Screenshots (optional)

Example with the mini.vim being used as a vimrc. Notice the highlighting around the (method)...
Screenshot 2019-10-09 at 09 28 02

A couple other snapshots with a theme to elaborate on what I mean about how it can completely throw off the way it looks. The first example is JS and the second is Scala

Screenshot 2019-10-08 at 14 13 20

Screenshot 2019-10-08 at 14 13 41

## Recap

I can't figure out why the () in the floating windows are being highlighted, and seemingly at times also what is between ` : ` in a type signature. How do I either turn this off or what is actually causing this?

oblitum:
NeoVim-only winhighlight feature is very prone to "misuse", as it essentially breaks Vim's highlighting linking system, making it hard to be ever employed with success.

winhl overrides accept built-in highlighting groups only, but even if that's followed (set winhl=Normal:MyGroup) to get a floating window's main text changed, one may get scrambled highlighting in case the floating windows contains text highlighted by custom groups linked to Normal (typescriptParens for example), these groups will still link to the original Normal highlight instead of the override, which equates to getting mixed "MyGroup Normal"/"Original Normal" content. Meaning, it leads to some parts of the text getting overridden highlight but other parts strangely not.

The current fix replaces occurrences for setting winhl with window local highlight links.

The highlight links are simple and directly forcibly created (without any highlight clear, etc.), this relies on a very clever feature of Vim highlighting links where having cycles isn't a problem at all, so there's no risk having NormalFloat linked to CocFloating which itself links to NormalFloat. Such kind of cycle ends up in having NormalFloat's original highlight, which is just what we want if the cycle happens.

Another solution instead of window local highlight exec's is to make use of syntax clusters, and it works, but as this requires an additional syntax file with the highlighting overrides to be loaded, I've dropped it.

Fixes #1251

neobones from [zenbones.nvim](https://github.com/mcchrish/zenbones.nvim):
img-2021-11-26-194208

[rose-pine](https://github.com/rose-pine/neovim):
img-2021-11-26-195143

### DIFF
--- a/autoload/coc/float.vim
+++ b/autoload/coc/float.vim
@@ -147,7 +147,12 @@ function! coc#float#create_float_win(winid, bufnr, config) abort
       return []
     endif
     let hlgroup = get(a:config, 'highlight', 'CocFloating')
-    call setwinvar(winid, '&winhl', 'Normal:'.hlgroup.',NormalNC:'.hlgroup.',FoldColumn:'.hlgroup)
+    call coc#compat#execute(winid, [
+          \ 'hi! link Normal '.hlgroup,
+          \ 'hi! link NormalNC '.hlgroup,
+          \ 'hi! link FoldColumn '.hlgroup,
+          \ 'hi! link NormalFloat '.hlgroup
+          \ ])
     call setwinvar(winid, '&signcolumn', 'no')
     call setwinvar(winid, '&foldenable', 0)
     " cursorline highlight not work on old neovim
@@ -257,14 +262,22 @@ function! coc#float#nvim_border_win(config, winid, border, title, hasbtn, hlgrou
   endif
   if winid
     call nvim_win_set_config(winid, opt)
-    call setwinvar(winid, '&winhl', 'Normal:'.a:hlgroup.',NormalNC:'.a:hlgroup)
+    call coc#compat#execute(winid, [
+          \ 'hi! link Normal '.a:hlgroup,
+          \ 'hi! link NormalNC '.a:hlgroup,
+          \ 'hi! link NormalFloat '.a:hlgroup
+          \ ])
   else
     noa let winid = nvim_open_win(bufnr, 0, opt)
     if winid
       if a:winblend
         call setwinvar(winid, '&winblend', a:winblend)
       endif
-      call setwinvar(winid, '&winhl', 'Normal:'.a:hlgroup.',NormalNC:'.a:hlgroup)
+      call coc#compat#execute(winid, [
+            \ 'hi! link Normal '.a:hlgroup,
+            \ 'hi! link NormalNC '.a:hlgroup,
+            \ 'hi! link NormalFloat '.a:hlgroup
+            \ ])
       call setwinvar(winid, 'target_winid', a:winid)
       call setwinvar(winid, 'kind', 'border')
       call add(a:related, winid)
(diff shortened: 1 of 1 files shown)


oblitum:
This fix contains a side-effect when the window is focused:

gif-2021-11-27-081529

Main window background gets the color of the float, but that's reset back when exiting the float.

Personally I don't mind this though as entering floats is quite rare thing to do in general, so it's much less worse to have this than bad highlighting.

Syntax clusters for highlighting the whole float could solve this issue completely, but it actually requires a full rewrite of floats around that, so it's too complex for me to handle atm.

oblitum:
This "fix" actually messes up `Normal` of non focused windows, it doesn't work right, back to drawing board 😭.

[side effect](https://user-images.githubusercontent.com/1269815/143722950-c5208cf4-70e9-4e91-a00b-30f3c9fd47e9.gif)

oblitum:
Seems syntax clusters is the only way out.

oblitum:
> ~~Seems syntax clusters is the only way out~~

The idea of "syntax cluster" was to roughly apply something like the following, by having multiple syntax files in the `contains=` of the problematic `syntax region`, the latter syntax file working as an override syntax file.

`autoload/coc/highlight.vim`
```diff
diff --git a/autoload/coc/highlight.vim b/autoload/coc/highlight.vim
index 1d77c566..ae3326ef 100644
--- a/autoload/coc/highlight.vim
+++ b/autoload/coc/highlight.vim
@@ -1,5 +1,6 @@
 scriptencoding utf-8
 let s:is_vim = !has('nvim')
+let s:root = expand('<sfile>:h:h:h')
 let s:clear_match_by_window = has('nvim-0.5.0') || has('patch-8.1.1084')
 let s:prop_offset = get(g:, 'coc_text_prop_offset', 1000)
 let s:namespace_map = {}
@@ -308,10 +309,11 @@ function! coc#highlight#highlight_lines(winid, blocks) abort
       endif
       if index(defined, filetype) == -1
         call add(cmds, 'syntax include @'.toupper(filetype).' syntax/'.filetype.'.vim')
+        call add(cmds, 'syntax include @OVERRIDE '.s:root.'/autoload/coc/override.vim')
         call add(cmds, 'unlet! b:current_syntax')
         call add(defined, filetype)
       endif
-      call add(cmds, 'syntax region CodeBlock'.region_id.' start=/\%'.start.'l/ end=/\%'.end.'l/ contains=@'.toupper(filetype).' keepend')
+      call add(cmds, 'syntax region CodeBlock'.region_id.' start=/\%'.start.'l/ end=/\%'.end.'l/ contains=@'.toupper(filetype).',@OVERRIDE'.' keepend')
       let region_id = region_id + 1
     endif
   endfor
```

`autoload/coc/override.vim`
```vim
highlight! link Normal      CocFloating
highlight! link NormalNC    CocFloating
highlight! link NormalFloat CocFloating
highlight! link FoldColumn  CocFloating
```

The premise was that the highlighting override would apply to the syntax region only, but it doesn't work like that, and it has the same side effect as commented previously.

So I give up :-/

Future nvim will probably have a fix when they deprecate `winhighlight`.

oblitum:
As @kevinhwang91 has suggested in gitter, configuring floats (see `:h coc-config-float`) for `Normal` highlight background can help hide the problem (both main and float will have same bg), to see the floats, enable border.

Issue #3388, opened 2021-09-20
CoC breaks on neovim's latest nightly


richin13:
## Result from CocInfo

While on neovim version *0.5.0*

```
## versions

vim version: NVIM v0.5.0
node version: v14.15.1
coc.nvim version: 0.0.80-03c9add7cd
coc.nvim directory: /home/rmadriz/.config/nvim/pack/packager/start/coc.nvim
term: alacritty
platform: linux

## Log of coc.nvim

2021-09-20T08:14:49.309 INFO (pid:122330) [services] - registered service "jedi"
2021-09-20T08:14:49.312 INFO (pid:122330) [services] - jedi-language-server state change: stopped => starting
2021-09-20T08:14:49.334 INFO (pid:122330) [services] - registered service "highlight"
2021-09-20T08:14:49.349 INFO (pid:122330) [services] - registered service "diagnostic-languageserver"
2021-09-20T08:14:49.349 INFO (pid:122330) [services] - diagnostic language service state change: stopped => starting
2021-09-20T08:14:49.351 INFO (pid:122330) [plugin] - coc.nvim 0.0.80-03c9add7cd initialized with node: v14.15.1 after 541ms
2021-09-20T08:14:49.356 INFO (pid:122330) [language-client-index] - Language server "jedi" started with 122527
2021-09-20T08:14:49.418 INFO (pid:122330) [attach] - receive notification: highlight []
2021-09-20T08:14:49.521 INFO (pid:122330) [services] - diagnostic language service state change: starting => running
2021-09-20T08:14:49.522 INFO (pid:122330) [services] - service diagnostic-languageserver started
2021-09-20T08:14:49.760 INFO (pid:122330) [services] - jedi-language-server state change: starting => running
2021-09-20T08:14:49.762 INFO (pid:122330) [services] - service jedi started
2021-09-20T08:14:53.568 INFO (pid:122330) [attach] - receive notification: rename []
2021-09-20T08:14:54.974 WARN (pid:122330) [plugin] - Slow action "rename" cost 1405ms
2021-09-20T08:14:55.336 INFO (pid:122330) [attach] - receive notification: highlight []
2021-09-20T08:14:57.029 INFO (pid:122330) [attach] - receive notification: showInfo []


```

Latest nightly build (NVIM v0.6.0-dev+304-g924e8e4f2)

```
## versions

vim version: NVIM v0.6.0-dev+304-g924e8e4f2
node version: v14.15.1
[...]

PR #3511, opened 2021-12-15
fix: correctly handle globs in 'runtimepath'

midchildan:
# The Issue
Neovim 0.6.0 has started to use globs in its default 'runtimepath' [1]. This prevented CoC from detecting extensions that are installed in `~/.config/nvim/pack/*/start/*`.

[1]: https://github.com/neovim/neovim/commit/9df7e02

# Results
I compared the output of `:CocList extensions` before and after applying the fix.

Before:
before

After:
Screen Shot 2021-12-16 at 2 07 47 AM

### DIFF
--- a/autoload/coc/api.vim
+++ b/autoload/coc/api.vim
@@ -149,7 +149,7 @@ function! s:funcs.feedkeys(keys, mode, escape_csi)
 endfunction
 
 function! s:funcs.list_runtime_paths()
-  return split(&runtimepath, ',')
+  return globpath(&runtimepath, '', 0, 1)
 endfunction
 
 function! s:funcs.command_output(cmd)
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -540,7 +540,7 @@ function! coc#util#vim_info()
         \ 'colorscheme': get(g:, 'colors_name', ''),
         \ 'workspaceFolders': get(g:, 'WorkspaceFolders', v:null),
         \ 'background': &background,
-        \ 'runtimepath': &runtimepath,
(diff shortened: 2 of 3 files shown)


fannheyward:
Does this work with vim native package manager?

midchildan:
Yes, it should work with Vim's native package manager. It's what I used to test the fix:

```console
❯ ls ~/.config/nvim/pack/home-manager/start
__python3_dependencies/ coc-go@                 coc-rust-analyzer@      coc-ultisnips@
coc-clangd@             coc-json@               coc-snippets@           coc.nvim@
coc-cmake@              coc-pyright@            coc-tsserver@           fzf@
```

I also confirmed that the fix also works for packages added with `:packadd`.

fannheyward:
cc @kevinhwang91 @oblitum sorry to boring you guys, I know you're using nvim packer system to manage plugins, can you review and test this PR? Does it fix https://github.com/neoclide/coc.nvim/issues/3388?

kevinhwang91:
Not enough, coc.nvim can't load RTP extensions under `pack/*/opt/{name}` correctly, even although `CocList extensions` show users the expected result.

The culprit comes from: https://github.com/neoclide/coc.nvim/blob/4ffab48b39b1e12e7dde7aef6dc256a03cb848a3/src/workspace.ts#L207

`fastDiff` can't handle the result fired by `packadd coc-xxxx` (I only tested under `coc-pyright`) correctly.
 
BTW, I think `globpath` is overkill to find the RTP extensions.

fannheyward (approved):

PR #3512, opened 2021-12-16
funcs.buf_clear_namespace should treat endLine as exclusive

crccw:
I noticed that diagnostic messages cancel the semantic highlight on the next line. I haven't figured out why diagnostic affects semantic highlighting, but it seems that funcs.buf_clear_namespace is wrongly treating endLine as inclusive (VIM textprop ranges are 1-based end-inclusive, while LSP ranges are 0-based end-exclusive).

buf_clear_namespace  is used by

- coc#highlight#update_highlights
- coc#highlight#clear_highlight

Both seems to accept end-exclusive ranges.


chemzqm:
Thank you.

PR #3515, opened 2021-12-18
Add rootPatterns to LanguageClient options

blackheaven:
I'm trying to create and extension for `haskell-language-server`, and I need to set `rootPatterns`.


chemzqm:
You should use `this._clientOptions.workspaceFolder` instead, which is recommended.
Use configuration `contributes.rootPatterns` for workspace folder resolved checkout https://github.com/neoclide/coc-tsserver/blob/master/package.json#L41 for example.

Issue #1940, opened 2020-05-17
[help] Auto complete floating window doesn't disappear when pressing ESC


waldemarnt:
Looks like my auto-complete floating window only disappear when I type something if I press ESC, for example, it stays there forever.

![preview](https://i.ibb.co/71pCBBG/Screen-Shot-2020-05-17-at-3-56-26-pm.png)

**Environment**
vim version: NVIM v0.5.0-5f92093
node version: v12.16.3
coc.nvim version: 0.0.78-d83536aa4e
term: iTerm.app
platform: darwin

[my .vimrc](https://github.com/waldemarnt/dotfiles/blob/master/.vimrc)

Any idea how can I make it hide for when pressing sec?

waldemarnt:
Just found out that it has been discussed here already https://github.com/neoclide/coc.nvim/issues/531 and looks like it's happening to more people, in the meantime I mapped `:call coc#util#float_hide()` to my ESC

chemzqm:
It should be hidden on CompleteDone

waldemarnt:
thansk @chemzqm is it a coc config? can you point me to an example please?

chemzqm:
Fixed on latest release branch of coc.nvim.

a-torgovitsky:
I am still encountering this problem occasionally. I am not sure how to reproduce it reliably though...
Not very helpful, sorry.

chemzqm:
Upgrade to latest release branch of coc.nvim

a-torgovitsky:
I was already pretty current. Just updated again and these are the new commits I was missing:
```vim
* f143e53 fix(handler): fix filetype of signature help (5 hours ago)
* 1d2bf0c chore(package): improve package.json for release branch (5 hours ago)
* fe2f462 fix(util): check winid for coc#util#close_win (16 hours ago)
* 6b7767f fix(workspace): avoid neovim's address error with bd! (28 hours ago)
```
is this one related?
```vim
* fe2f462 fix(util): check winid for coc#util#close_win (16 hours ago)
```

chemzqm:
Try upgrade your vim or provide reproduce way.

a-torgovitsky:
nvim v0.4.3

I will let you know if I find a way to reproduce
So far I haven't been able to figure out the pattern

PR #3516, opened 2021-12-20
fix(completion): keep floatwin from redrawing

kevinhwang91:
Look like `this.floating.close()` be added after https://github.com/neoclide/coc.nvim/issues/1940. However, I can't reproduce the issue.  Let's use debounce to do defensive programming.

Every call `complete()` will fire `CompleteDone` event.
The fired function will close the floating window at once and create a new one for the suggestion.

Solution: use debounce to close the floating window.

### DIFF
--- a/src/completion/index.ts
+++ b/src/completion/index.ts
@@ -37,6 +37,7 @@ export class Completion implements Disposable {
   private insertCharTs = 0
   private insertLeaveTs = 0
   private excludeImages: boolean
+  private closeFloat: Function & { clear(): void }
 
   public init(): void {
     this.config = this.getCompleteConfig()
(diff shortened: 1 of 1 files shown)


src/completion/index.ts
```diff
@@ -68,17 +69,24 @@ export class Completion implements Disposable {
     events.on('TextChangedP', this.onTextChangedP, this, this.disposables)
     events.on('TextChangedI', this.onTextChangedI, this, this.disposables)
     let fn = debounce(this.onPumChange.bind(this), 20)
+    this.closeFloat = debounce(async () => {
+      let visible = await this.nvim.call('pumvisible', []) as number
```
chemzqm, line 73: The pum could be created by other vim plugin, and the floating window should be closed in that case as well.

chemzqm (changes requested):

src/completion/index.ts
kevinhwang91: I'm not sure which issue makes the floating window can't be closed. Can you give me a tip for that? How about just calling `this.stop` side debounce?

chemzqm: Should be issue with old vim version, you can just remove `this.floating.close()` in `CompleteDone` event handler, should works fine.

chemzqm: Found bug with `<C-e>` after `<C-p>`, the completion not stopped.

kevinhwang91: Yes, I can reproduce it without this PR. It works fine with `this.stop` with debounce.

chemzqm:
Should be fixed by 7f5cf023

kevinhwang91:
Nice!

PR #3520, opened 2021-12-23
feat(model): add document.lineAt function

fannheyward:


chemzqm:
Need change for `typings/index.d.ts` as well

chemzqm (changes requested):
Still missing changes in typings/index.d.ts and tests for that function.

chemzqm:
Need export `TextLIne` class in index.ts, I'll fix that.

PR #3528, opened 2021-12-27
Feat(rename.ts): Tweak To Refactor-Rename Floating Window

cpearce-py:
Hi there all! 

I thought I'd share a small tweak that I've enjoyed having implemented in my own setup for coc.
When renaming, I find it more "natural" for typing to have the user input window display without the previous name. This removes the need to hold delete to remove the old name, before entering a new one.
I decided to add a conditional statement to check for a new config key,`coc.preferences.fillRename`, to decide whether to pass a default value to `window.requestInput()`. 
Default is true, which keeps the functionality the same as currently. False means the floating window is empty and the user can immediately enter a new name!
Would love to hear what you think, and sorry in advanced if anything isn't setup correctly, this is my first official PR for a project!

Many thanks:)

### DIFF
--- a/doc/coc.txt
+++ b/doc/coc.txt
@@ -916,6 +916,12 @@ Built-in configurations:~
 
 	Note: this configuration is expected to change in the future.
 
+"coc.preferences.fillRename"~
+
+	Disable to stop Refactor-Rename float/popup window from populating
+	with old name in the New Name field.
+	Default: `true`
+
 						*coc-config-cursors*
 
 "cursors.cancelKey":~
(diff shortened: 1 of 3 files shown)


doc/coc.txt
```diff
@@ -916,6 +916,12 @@ Built-in configurations:~
 
 	Note: this configuration is expected to change in the future.
 
+"coc.preferences.fillRename"~
```
chemzqm, line 919: My suggestion is change the name to `coc.preferences.renameFillCurrent`

chemzqm (changes requested):

doc/coc.txt
cpearce-py: Yes, I wasn't too sure if what I'd chosen was the best name. Have pushed the changes now.

PR #3529, opened 2021-12-27
perf(workspace): speed up converting locations to quickfix list

kevinhwang91:
Location may get the `text` of quickfix tiem by reading file content.
Even if language server has indexed files, it will be slow to read a
large number of duplicate files again and again.

Solution: use Mutex to read each file and cache read content.


kevinhwang91:
I have tried to sort the `range.start.line` and use `export async function readFileLines(fullpath: string, start: number, end: number): Promise<string[]>
` read lines by line range (have change `fs.existsSync` to async `statAsync` inside).
Reading file content totally is faster than reading file by line range, just needs more memory to use.

chemzqm (changes requested):
I don't think you need `mutex` and we have mutex support in `src/util/mutex.ts`

PR #3532, opened 2021-12-28
fix(document): document did changed event could be lost

xuanduc987:
If we config coc to format typescript file with prettier on save

```ts
let x = 1
```

type `o<esc>ix<esc>:w<cr>` quickly

-> the last line got deleted.


xuanduc987:
This problem occurs frequently for me, because I have the following mapping
```
nnoremap <leader>w :update<CR>
```
thus I could quickly save file right after edit.

chemzqm:
Thanks, I've made a fix with 4a1cc23b

PR #3550, opened 2022-01-06
fix(doc): Typos

cincodenada:
Just fixing a few typos here and there - I ran across "chagnes" and then
just went through the whole file with a spellcheck


cincodenada:
I wasn't sure if this should be fix(doc) or docs(doc) - looking through history, it looks like typos are mostly tagged as fix(doc), but please edit the title if docs(doc) is preferred!

fannheyward (approved):

PR #3601, opened 2022-01-25
perf(semanticTokensHighlights): disable onDidSemanticTokensRefresh

kevinhwang91:
The server like clangd will fire onDidSemanticTokensRefresh, but
SemanticTokensRefresh only provide selector instead of document, if we
request semanticTokens concurrently, we will encounter performance
issue.

On the server side:
Multiple semanticTokens request would make completion request lagging.

On the client side(coc.nvim):
After getting the semanticTokens results without cancellation, converting
token to highlights eating nodejs resources and diffHighlight may fail
and then make Neovim 100% cpu usage.

Look like we have enough events to request semanticTokens, no need to
enable onDidSemanticTokensRefresh.

### DIFF
--- a/src/handler/semanticTokensHighlights/index.ts
+++ b/src/handler/semanticTokensHighlights/index.ts
@@ -65,14 +65,14 @@ export default class SemanticTokensHighlights {
     this.highlighters = workspace.registerBufferSync(doc => {
       return new SemanticTokensBuffer(this.nvim, doc.bufnr, this.config)
     })
-    languages.onDidSemanticTokensRefresh(selector => {
-      for (let item of this.highlighters.items) {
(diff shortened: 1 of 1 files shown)


chemzqm:
It's required to refresh semantic tokens.

chemzqm:
> This is useful if a server detects a project wide configuration change which requires a re-calculation of all semantic tokens. 

Seems issue with clangd which should not send refresh request too frequently.

kevinhwang91:
If I open 10 c filetype buffers, only one onDidSemanticTokensRefresh event will make nodejs send ten requests for SemanticTokens concurrently.

chemzqm:
I've improved it d0a764e5

PR #3621, opened 2022-02-05
[WIP] Use lua functions on neovim >= 0.6.0 for better performance

chemzqm:
The lua functions could be much faster, could help to reduce the block on neovim.


kevinhwang91:
Improved a lot compare with Vimscript in a large file.

Call Lua from Typescript directly should be better performance.
```typescript
et curr = await this.nvim.lua(`return require('coc').getHighlights(...)`, [bufnr, ns]) as HighlightItemResult[]
```

Waiting for this PR to be merged.

PR #3634, opened 2022-02-10
fix(line-selection) coc#cursor#get_selection

whyreal:


chemzqm:
It's line mode, we should include line ending.

whyreal:
Yes, we should include line ending, but the next line after the range should not be included, Right!?

PR #3636, opened 2022-02-12
perf(semanticTokens): avoid parsing all returned data

kevinhwang91:
There're several performance bottlenecks for the semantic token in all projects/files on client side:

Just take Neovim as an example, Vim is similar.

The data returned by the server will be all parsed and asked for total extmarks from Neovim.
If the data is large:
1. coc.nvim will be lagging to handle other services asked from Neovim;
2. Neovim must fetch all extmarks which may be blocking a few hundred milliseconds;
3. Neovim must encode the data also may be blocking;

Solution From this PR:
1. Only parse the relative semantic tokens format to the absolute format.
   This phase is a very small cost for coc.nvim;
2. Centered on the cursor, only render 3 times window height range from the absolute format
   pending semantic tokens, Neovim can understand the line number. 1 window
   height range is not enough, users who move up and down can feel the
   redraw frequently. Scroll the half-page down (`<C-d>`) on the topline
   will span 1.5 times window height range, scroll the half-page up is similar.
   In other words, to achieve less redraw must render 3 times window height
   if users press `<C-d>` or `<C-u>`;
3. Only diff 3 times window height range between semantic token and extmarks;

### DIFF
--- a/src/handler/semanticTokensHighlights/buffer.ts
+++ b/src/handler/semanticTokensHighlights/buffer.ts
@@ -353,17 +353,11 @@ export default class SemanticTokensBuffer implements SyncItem {
       items = await this.requestAllHighlights(startLine, endLine, tokenSource.token, forceFull)
     }
     // request cancelled or can't work
-    if (!items || tokenSource.token.isCancellationRequested || this.invalid) {
-      this.tokenSource = null
-      return
-    }
+    if (!items || tokenSource.token.isCancellationRequested || this.invalid) return
     let diff = await window.diffHighlights(this.bufnr, NAMESPACE, items, startLine, endLine)
-    if (tokenSource.token.isCancellationRequested || !diff || this.invalid) {
-      this.tokenSource = null
-      return
-    }
-    await window.applyDiffHighlights(this.bufnr, NAMESPACE, priority, diff)
     this.tokenSource = null
+    if (tokenSource.token.isCancellationRequested || !diff || this.invalid) return
+    await window.applyDiffHighlights(this.bufnr, NAMESPACE, priority, diff)
   }
 
   /**
--- a/src/handler/semanticTokensHighlights/buffer.ts
+++ b/src/handler/semanticTokensHighlights/buffer.ts
@@ -123,7 +123,7 @@ export default class SemanticTokensBuffer implements SyncItem {
 
   public get shouldRangeHighlight(): boolean {
     let doc = workspace.getDocument(this.bufnr)
-    if (!doc) return false
+    if (!doc || this.tokenSource) return false
     return languages.hasProvider('semanticTokensRange', doc.textDocument) && this.previousResults == null
   }
 
@@ -284,7 +284,7 @@ export default class SemanticTokensBuffer implements SyncItem {
   }
 
   public async doPendingHighlight() {
-    if (this._pendingHighlights.size <= 0) return
+    if (!this.hasPendingHighlights) return
     let r = await this.nvim.call('coc#window#render_range', [this.bufnr]) as [number, number]
     if (!r) return
     let [startLine, endLine] = r
@@ -332,7 +332,6 @@ export default class SemanticTokensBuffer implements SyncItem {
     let { nvim } = this
     let visible = await nvim.call('pumvisible')
     if (visible) return
-    let tokenSource = this.tokenSource = new CancellationTokenSource()
     let r = await nvim.call('coc#window#render_range', [this.bufnr]) as [number, number]
     if (!r) return
     let [startLine, endLine] = r
@@ -344,6 +343,7 @@ export default class SemanticTokensBuffer implements SyncItem {
     let previousVersion = this.previousResults?.version
     let doc = workspace.getDocument(this.bufnr)
     let items: HighlightItem[] | undefined
+    let tokenSource = this.tokenSource = new CancellationTokenSource()
     // TextDocument not changed, need perform highlight since lines possible changed.
     if (previousVersion && doc && doc.version === previousVersion) {
       let tokens = this.previousResults.tokens
@@ -353,11 +353,17 @@ export default class SemanticTokensBuffer implements SyncItem {
       items = await this.requestAllHighlights(startLine, endLine, tokenSource.token, forceFull)
     }
     // request cancelled or can't work
-    if (!items || tokenSource.token.isCancellationRequested || this.invalid) return
+    if (!items || tokenSource.token.isCancellationRequested || this.invalid) {
+      this.tokenSource = null
+      return
+    }
     let diff = await window.diffHighlights(this.bufnr, NAMESPACE, items, startLine, endLine)
-    this.tokenSource = null
-    if (tokenSource.token.isCancellationRequested || !diff || this.invalid) return
+    if (tokenSource.token.isCancellationRequested || !diff || this.invalid) {
+      this.tokenSource = null
+      return
+    }
     await window.applyDiffHighlights(this.bufnr, NAMESPACE, priority, diff)
+    this.tokenSource = null
   }
 
   /**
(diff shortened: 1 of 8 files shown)


chemzqm:
How to reproduce the issue?

kevinhwang91:
> How to reproduce the issue?

Do you mean the performance issue? Enable semantic token for `c` and open `https://github.com/neovim/neovim/blob/master/src/nvim/eval.c`, doing some editing and undo, move the cursor, you will feel the lagging for Neovim. Especially, doing some global change like `Gread` in fugitive is more easily perceptible.

kevinhwang91:
@chemzqm @fannheyward @oblitum I'm sorry to bother you. Have you tried this PR? Editing text or scrolling screen is smooth.

kevinhwang91:
> @kevinhwang91 I can't reproduce lag using clangd with `src/nvim/eval.c`, could you try add some logs? For example:
> 
> ```ts
>     let d = Date.now()
>     if (SemanticTokens.is(result)) {
>       tokens = result.data
>     } else if (previousResult && Array.isArray(result.edits)) {
>       tokens = previousResult.tokens
>       result.edits.forEach(e => {
>         tokens.splice(e.start, e.deleteCount ? e.deleteCount : 0, ...e.data)
>       })
>     } else {
>       return
>     }
>     this.previousResults = { resultId: result.resultId, tokens, version }
>     let res = this.getHighlightItems(doc, tokens, legend)
>     logger.debug('len:', tokens.length, res.length)
>     logger.debug('cost:', Date.now() - d)
> ```

The snippet you provided is a small cost.

https://github.com/neoclide/coc.nvim/blob/72aeffd7abb9dda236e973a404e99e36c7c29a63/src/handler/semanticTokensHighlights/buffer.ts#L174-L193 cost nodejs nearly 40-80ms

https://github.com/neoclide/coc.nvim/blob/72aeffd7abb9dda236e973a404e99e36c7c29a63/src/window.ts#L551 block Neovim 150-300ms (without https://github.com/neoclide/coc.nvim/pull/3621)

chemzqm:
> cost nodejs nearly 40-80ms

Don't know why your computer so slow, it's about 4~5ms on my computer.

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -255,54 +329,61 @@ export default class SemanticTokensBuffer implements SyncItem {
   public async doHighlight(forceFull = false): Promise<void> {
     if (!this.enabled) return
     this.cancel()
-    let visible = await this.nvim.call('pumvisible')
+    let { nvim } = this
+    let visible = await nvim.call('pumvisible')
     if (visible) return
-    let tokenSource = this.tokenSource = new CancellationTokenSource()
+    let r = await nvim.call('coc#window#render_range', [this.bufnr]) as [number, number]
```
chemzqm, line 335: The range could be changed (on CursorMoved) during request, so you'd better not request the range here.

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -255,54 +329,61 @@ export default class SemanticTokensBuffer implements SyncItem {
...
     // request cancelled or can't work
-    if (!items || tokenSource.token.isCancellationRequested || this.invalid) return
-    let diff = await window.diffHighlights(this.bufnr, NAMESPACE, items)
+    // shouldRangeHighlight need check this.tokenSource
+    if (!items || tokenSource.token.isCancellationRequested || this.invalid) {
+      tokenSource.cancel()
+      this.tokenSource = null
+      return
+    }
+    let diff = await window.diffHighlights(this.bufnr, NAMESPACE, items, startLine, endLine)
```
chemzqm, line 362: You can simply use `coc#highlight#update_highlights` for range update instead diff and apply, which would be faster. Further more, it's better to perform diff apply when the document not have any highlights applied (`coc#highlight#get_highlights` would not cause delay without highlight items).

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -240,12 +283,43 @@ export default class SemanticTokensBuffer implements SyncItem {
...
+    }
+  }
+
+  public async doRangeHighlight(startLine?: number, endLine?: number): Promise<void> {
     if (!this.enabled) return
+    if (!startLine || !endLine) {
+      let r = await this.nvim.call('coc#window#render_range', [this.bufnr]) as [number, number]
+      if (!r) return
+      [startLine, endLine] = r
+    }
```
chemzqm, line 318: Not looks good since it's not guard by CancellationTokenSource any more.

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -240,12 +283,43 @@ export default class SemanticTokensBuffer implements SyncItem {
...
+    let r = await this.nvim.call('coc#window#render_range', [this.bufnr]) as [number, number]
+    if (!r) return
+    let [startLine, endLine] = r
+    const items: HighlightItem[] = []
+    let doc = workspace.getDocument(this.bufnr)
+    let start: number | undefined
+    let end: number | undefined
+    for (let i = startLine - 1; i < endLine; i++) {
+      let arr = this._pendingHighlights.get(i)
+      if (!this._pendingHighlights.delete(i) || arr.length == 0) continue
```
chemzqm, line 297: This delete could make further `coc#highlight#update_highlights` bug since it requires all highlight items in that range, but you may lost them by delete on previous `doPendingHighlight `

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -124,7 +140,21 @@ export default class SemanticTokensBuffer implements SyncItem {
    * Get current highlight items
    */
   public get highlights(): ReadonlyArray<SemanticTokenRange> {
-    return this._highlights
+    let pending: SemanticTokenRange[] = []
+    this._pendingHighlights.forEach(absTokens => {
```
chemzqm, line 144: `semanticTokens.inspect` requires all current highlights, but `_pendingHighlights` changes from time to time.

chemzqm (commented):
你先别弄了，我这里重新做些优化吧

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -353,7 +353,12 @@ export default class SemanticTokensBuffer implements SyncItem {
       items = await this.requestAllHighlights(startLine, endLine, tokenSource.token, forceFull)
     }
     // request cancelled or can't work
-    if (!items || tokenSource.token.isCancellationRequested || this.invalid) return
+    // shouldRangeHighlight need check this.tokenSource
+    if (!items || tokenSource.token.isCancellationRequested || this.invalid) {
+      tokenSource.cancel()
+      this.tokenSource = null
```
chemzqm, line 359: You can't simply reset `this.tokenSource` since it could be changed during request

src/handler/semanticTokensHighlights/buffer.ts
```diff
@@ -111,10 +123,14 @@ export default class SemanticTokensBuffer implements SyncItem {
 
   public get shouldRangeHighlight(): boolean {
     let doc = workspace.getDocument(this.bufnr)
-    if (!doc) return false
+    if (!doc || this.tokenSource) return false
```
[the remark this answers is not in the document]
kevinhwang91: I don't think 

> 你先别弄了，我这里重新做些优化吧

ok

kevinhwang91:
> 你先别弄了，我这里重新做些优化吧

好

src/handler/semanticTokensHighlights/buffer.ts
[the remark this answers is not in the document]
chemzqm: `requestAllHighlights` could be canceled or return `null`, range highlight should not be canceled on that case.

PR #3656, opened 2022-02-18
fix(highlight): enable right_gravity for extmark

kevinhwang91:
If `right_gravity` = false, inserting new line above `[row, 0, row, end_col]` extmark will
extend unexpected range which become [row - 1, 0, row, end_col],
and delta will be great than 0.

Press `[v:count]O` under one extmark that range is `[row, 0, row, end_col]` can easy to reproduce the issue.

### DIFF
--- a/autoload/coc/highlight.vim
+++ b/autoload/coc/highlight.vim
@@ -353,7 +353,7 @@ function! coc#highlight#add_highlight(bufnr, src_id, hl_group, line, col_start, 
               \ 'end_col': a:col_end,
               \ 'hl_group': a:hl_group,
               \ 'hl_mode': get(opts, 'combine', 1) ? 'combine' : 'replace',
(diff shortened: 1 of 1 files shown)


chemzqm:
Seems neovim have issue when `right_gravity` enabled, https://github.com/neovim/neovim/issues/17170

kevinhwang91:
Have you tried to disable `right_gravity` except the `[row, 0]` extmark as a workaround? I am not easy to reproduce the upstream issue.

chemzqm:
No, can't reproduce your issue, please provide your neovim version.

kevinhwang91:
`v0.6.1` or `master`
```go
package main

func main() {
}
```

Press `O` under `func main()` and then type some chars.

chemzqm:
I think that neovim should disable increase the highlight when user type none keyword character, as what vim does.

PR #3693, opened 2022-03-10
fix(semanticTokens): don't abandon highlight for region buffer

kevinhwang91:
For now, onDidSemanticTokensRefresh will abandon highlight for buffers
which highlighted by region.


chemzqm:
Why?

PR #3699, opened 2022-03-13
feat(commands): add `editor.action.formatDocument` command

xiyaowong:


fannheyward:
Why not add this to `src/commands.ts` with something like `await plugin.cocAction('format')`?

src/handler/format.ts
```diff
@@ -83,6 +84,17 @@ export default class FormatHandler {
       let pre = info.pre[info.pre.length - 1]
       if (pre) await this.tryFormatOnType(pre, bufnr)
     }))
+
+    handler.addDisposable(commandManager.registerCommand('editor.action.formatDocument', async (bufnr?: number) => {
```
chemzqm, line 88: It's better to use `uri` instead of `bufnr`.

chemzqm (changes requested):

src/handler/format.ts
xiyaowong: Why? Users are more likely to provide bufnr, same as organizeImport, although the most of users may not know that arguments can be provided

chemzqm: Because one `bufnr` could be used for different document, try `:tabe | echo bufnr('%')` and then open another file, you can see they share same bufnr.

xiyaowong: I don't understand. And following your steps, `bufnr` has not been the same.

kevinhwang91: IMO, coc.nvim is LSP-related framework, users should translate bufnr to uri. Please run `:h uri_from_bufnr` In Neovim.

xiyaowong: thx, I have known this section. Just don't understand many other methods are used bufnr, this is a historical legacy or other considerations?

chemzqm: The buffer could be not attached for some reasons `:h CocAction('ensureDocument')`, using `uri` would be safer.

chemzqm: You can use `bufnr` or `uri`, `bufnr` is for vim staff and `uri` for LSP related.

PR #3703, opened 2022-03-14
fix(completion): use nvim_select_popupmenu_item to close pum if avaliable

xiyaowong:


chemzqm:
The doc of `nvim_select_popupmenu_item` said `Value of -1 selects nothing and restores the original text.`, it's not expected since we expect to close the pum only.
I've tried `nvim_select_popupmenu_item` and it doesn't restart the original text, seems bug with neoivm.

xiyaowong:
Good to know

xiyaowong:
I think it would be better if nvim could provide an option to determine these behaviors

PR #3706, opened 2022-03-15
fix(diagnostic): if calls diagnosticRefresh, should refresh in any mode

xiyaowong:


chemzqm:
Why you need refresh diagnostics on insert mode?

xiyaowong:
This method is for the user, the check of the current mode should be left to the user.

PR #3712, opened 2022-03-17
fix(diagnostic): respect priority of highlight groups

kevinhwang91:
No idea how to write a test to get the current diagnostic highlight showed in Neovim.

### DIFF
--- a/autoload/coc/highlight.vim
+++ b/autoload/coc/highlight.vim
@@ -355,7 +355,7 @@ function! coc#highlight#add_highlight(bufnr, src_id, hl_group, line, col_start, 
               \ 'hl_mode': get(opts, 'combine', 1) ? 'combine' : 'replace',
               \ 'right_gravity': v:true,
               \ 'end_right_gravity': v:false,
-              \ 'priority': type(priority) == 0 ?  min([priority, 4096]) : 4096,
+              \ 'priority': type(priority) == 0 ?  max([priority, 101]) : 4096,
               \ })
       catch /^Vim\%((\a\+)\)\=:E5555/
         " the end_col could be invalid, ignore this error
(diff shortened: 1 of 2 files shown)


kevinhwang91:
I found that I misunderstood the `updateHighlights` method, need fix.

kevinhwang91:
Could we add a high-level API to add highlights with priority in batch? Look like there's only a vimscript `coc#highlight#set` function, for now.

autoload/coc/highlight.vim
```diff
@@ -355,7 +355,7 @@ function! coc#highlight#add_highlight(bufnr, src_id, hl_group, line, col_start,
               \ 'hl_mode': get(opts, 'combine', 1) ? 'combine' : 'replace',
               \ 'right_gravity': v:true,
               \ 'end_right_gravity': v:false,
-              \ 'priority': type(priority) == 0 ?  min([priority, 4096]) : 4096,
+              \ 'priority': type(priority) == 0 ?  max([priority, 101]) : 4096,
```
chemzqm, line 358: You should not break exist check

src/diagnostic/buffer.ts
```diff
@@ -334,9 +334,22 @@ export class DiagnosticBuffer implements SyncItem {
...
+        return p
+      }, new Map<string, HighlightItemDef[]>())
+      let highlightPriority = this.config.highlightPriority
+      nvim.pauseNotification()
+      this.clearHighlight(collection)
+      for (let [hlGroup, items] of grouped) {
+        let priority = highlightPriority + hlGroups.length - hlGroups.indexOf(hlGroup)
+        nvim.call('coc#highlight#set', [this.bufnr, NAMESPACE + collection, items, priority], true)
+      }
+      void nvim.resumeNotification(true, true)
```
chemzqm, line 352: It causes nested `nvim.pauseNotification()` which can be buggy and unnecessary redraw on vim.

PR #3713, opened 2022-03-17
feat: Compatible with the extension host class of Vscode

weirongxu:
I've try to port some vscode extensions. (like [mintlify](https://github.com/mintlify/vscode-docs))
However, this extension uses the `MarkdownString` API, so it's a bit difficult to port to coc.
So I tried to migrate some extension host classes to coc.

vscode source https://github.com/microsoft/vscode/blob/main/src/vs/workbench/api/common/extHostTypes.ts

Regarding the `vscode-languageserver-protocol` API (e.g. `Range.create()`, which can be replaced by `new Range()`).
I'm not sure whether we are keeping these older APIs, I can mark it as **@deprecated**, but another option is to keep them backward compatible.
Please let me know if you have any suggestions about this.

## TODO
- [ ] classes
  - [x] Position
  - [x] Range
  - [x] Selection
  - [x] TextEdit
  - [ ] WorkspaceEdit (pending)
  - [x] Location
  - [ ] ~~Diagnostic~~
  - [ ] ~~Hover~~
  - [x] MarkdownString
- [ ] export
- [ ] typing
- [ ] unit test

### DIFF
--- a/src/edit/location.ts
+++ b/src/edit/location.ts
@@ -16,7 +16,7 @@ export class Location implements ILocation {
       return false
     }
     return Range.isRange((thing as Location).range)
-    && URI.isUri((thing as Location).uri)
+      && URI.isUri((thing as Location).uri)
   }
 
   /**
--- a/src/edit/location.ts
+++ b/src/edit/location.ts
@@ -1,0 +1,64 @@
+/* ---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { Location as ILocation } from 'vscode-languageserver-protocol'
+import { URI } from 'vscode-uri'
+import { Position } from './position'
+import { Range } from './range'
+
+export class Location implements ILocation {
+  public static isLocation(thing: any): thing is ILocation {
+    if (thing instanceof Location) {
+      return true
+    }
+    if (!thing) {
+      return false
+    }
+    return Range.isRange((thing as Location).range)
+    && URI.isUri((thing as Location).uri)
+  }
+
+  /**
+   * Creates a Location literal.
+   *
+   * @param uri The location's uri.
+   * @param range The location's range.
+   * @deprecated use `new Location(uri, range)` instead.
+   */
+  public static create(uri: string, range: Range): Location {
+    return new Location(uri, range)
+  }
+  /**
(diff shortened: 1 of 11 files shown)


chemzqm:
It seems you need rename those class names to something like `VPosition`, since coc.nvim already exposed them from `vscode-languageserver-types`

weirongxu:
> It seems you need rename those class names to something like `VPosition`, since coc.nvim already exposed them from `vscode-languageserver-types`

Thank you for your reply.
Can we keep the `Position` of the vscode host compatible with the `VPosition`? In this case, just need to expose the new Postion.

I'm busy these days, I'll probably come back in a week

chemzqm:
> Can we keep the VPosition of the vscode host compatible with the Position? In this case, just need to expose the new Postion

The current `Position` is an `interface` and `namespace`, unless you only add static methods to exists namespace, it can't be compatible. Or rename all `Position` in typings.d.ts to something like `IPosition`

weirongxu:
> The current `Position` is an `interface` and `namespace`, unless you only add static methods to exists namespace, it can't be compatible. Or rename all `Position` in typings.d.ts to something like `IPosition`

I got it now, I'll make some effort, otherwise use a new name like `VPosition`, and thanks again

luisdavim:
is this abandoned?

chemzqm:
I prefer a new module named `coc-vscode-adaptor` which provide APIs the same as VSCode, which would make migrate VSCode extensions much easier.

weirongxu:
I'm really sorry, because I couldn't focus on the development of coc related extensions for a long time before, which caused this pr to be stranded.

I'll check the code and associated implementation again after a while. If I can't go on, we would use another extension to implement the compatibility layer as chemzqm said.

Freed-Wu:
Comes from #5235. any result or progress?

ChuYanLon:
Hi @weirongxu and @chemzqm, this PR is very relevant to our recent work on [**coc-vscode-loader**](https://github.com/coc-plugin/coc-vscode-loader) — a tool that automatically converts VS Code extensions to coc plugins through source-level transforms (import mapping, API polyfills, LanguageClient adaptation) and provides a TUI installer. Rather than adding VS Code extension host classes to coc.nvim core, we took the approach of a **converter + loader plugin** that adapts extensions at build time, which keeps coc.nvim itself lightweight. We now have 128+ registry entries covering LSP servers (clangd, rust-analyzer, pyright, ESLint, etc.), snippets, and direct-API extensions. The approach @chemzqm suggested (a separate adaptor module) aligns well with our design. Would love to hear your thoughts and explore synergies.

PR #3717, opened 2022-03-18
workspace ignore folders ignore

### LINKED ISSUE #3716: Adding glob pattern to `workspace.ignoreFolders` didn't work
## Result from CocInfo

Running on master

## Describe the bug

See test in PR: #3717 : https://github.com/neoclide/coc.nvim/pull/3717/files#diff-0a7867e64cfd99ae5f7a1db0d0042a9a3d236dd346ae048a5580703ac26a89b2

According to docs:

```
"workspace.ignoredFolders":~

	List of folders that should not be resolved as workspace folder.
	Environment variables and minimatch patterns can be used.

	Default: ["$HOME"]
```

A glob pattern specified in `workspace.ignoredFolders` should be indeed ignored, but it's still added as workspace folders

jackieli-tes:
fix #3716

### DIFF
--- a/src/util/fs.ts
+++ b/src/util/fs.ts
@@ -45,19 +45,27 @@ export async function isGitIgnored(fullpath: string): Promise<boolean> {
   return false
 }
 
+export function resolveRoot(folder: string, subs: string[], cwd?: string, bottomup = false, checkCwd = true, ignored: string[] = []): string | null {
+  let dir = resolveRootNoIgnore(folder, subs, cwd, bottomup, checkCwd)
+  if (isFolderIgnored(dir, ignored)) {
+    return null
+  }
+  return dir
+}
+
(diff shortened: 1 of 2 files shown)


src/util/fs.ts
```diff
@@ -45,19 +45,27 @@ export async function isGitIgnored(fullpath: string): Promise<boolean> {
   return false
 }
 
+export function resolveRoot(folder: string, subs: string[], cwd?: string, bottomup = false, checkCwd = true, ignored: string[] = []): string | null {
+  let dir = resolveRootNoIgnore(folder, subs, cwd, bottomup, checkCwd)
+  if (isFolderIgnored(dir, ignored)) {
```
chemzqm, line 50: It's not expected, consider if the current cwd is ignored and this will prevent check for other parent folders.

PR #3724, opened 2022-03-26
fix(sources): respect ignoreGitignore for buffer

fannheyward:
I came this issue when opening x.log in another buffer, words from x.log were provided by buffer source .

The *.log files should be ignored, I found this was removed in  abb0232824e1e66f5884afff9c074279bc1b19f1, `document.isIgnored` is always false now. Should we add this back? @chemzqm

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -38,7 +38,6 @@ export interface ChangeInfo {
 // getText, positionAt, offsetAt
 export default class Document {
   public buftype: string
-  public isIgnored = false
   public chars: Chars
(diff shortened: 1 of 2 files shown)


fannheyward:
There maybe some perf issue to do gitignore checking for multiple buffers, related https://github.com/neoclide/coc.nvim/issues/3575

chemzqm:
Fixed by 9b5f9ec4

chemzqm:
Fixed by 9b5f9ec4

PR #3735, opened 2022-03-31
feat: add "ignoreRegexps" for built-in sources

falamous:
Added a simple way to ignore suggestions for `around` and `buffer` sources. I did not find tests for a similar `ignorePatterns` feature for the `file` source, so I assumed they were not necessary. An example use case would be to ignore number suggestions with `"^[0-9]*$"` for writing code with a lot of random numbers like tests.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -1600,6 +1600,22 @@
         "type": "string"
       }
     },
+    "coc.source.around.ignoreRegexps": {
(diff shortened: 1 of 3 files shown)


chemzqm:
I think you should add a configuration like `suggest.ignoreRegexps` to ignore trigger input, instead to the sources.

PR #3748, opened 2022-04-06
feat(completion): add suggest.ignoreRegexps

fannheyward:
close #3735


src/completion/index.ts
```diff
@@ -338,6 +339,15 @@ export class Completion implements Disposable {
       option.input = this.getInput(doc, pre)
       option.col = byteLength(pre) - byteLength(option.input)
     }
+    if (this.config.ignoreRegexps.length) {
```
chemzqm, line 342: Should check `option.input.length > 0` here

chemzqm (changes requested):

PR #3752, opened 2022-04-06
feat(list): expose listItems for extensions

kevinhwang91:
Every extension can register a source for list, but other extensions
must call `CocAction('listLoadItems', name)` to get the list items by
name. However, some sources like `symbols` may return a ton of data.
For a extension to get data, Neo/Vim and nodeJS must encode & decode
for the data which is very inefficient.

Solution:
We can expose a internal command for the extension, which is
efficient and more friendly to pure Typescript.

### DIFF
--- a/src/list/manager.ts
+++ b/src/list/manager.ts
@@ -22,6 +22,7 @@ import OutlineList from './source/outline'
 import ServicesList from './source/services'
 import SourcesList from './source/sources'
 import SymbolsList from './source/symbols'
+import commands from '../commands'
 const logger = require('../util/logger')('list-manager')
 
 const mouseKeys = ['<LeftMouse>', '<LeftDrag>', '<LeftRelease>', '<2-LeftMouse>']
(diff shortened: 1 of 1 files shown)


kevinhwang91:
Look like some language servers can't return any data from `getWorkspaceSymbols` with an empty query. If you think this PR is a good idea, maybe should rewrite/overload `loadItems` in manager.ts.

chemzqm:
I don't think it's good idea to expose such kind of api, the list result could be changed from `ListItem[]` to `ListTask`
``` typescript
  loadItems(context: ListContext, token: CancellationToken): Promise<ListItem[] | ListTask | null | undefined>
```
if you need the result from a specific list, there could a specific command for that.

chemzqm:
I don't think it's good idea to expose such kind of api, the list result could be changed from `ListItem[]` to `ListTask`
``` typescript
  loadItems(context: ListContext, token: CancellationToken): Promise<ListItem[] | ListTask | null | undefined>
```
if you need the result from a specific list, there could a specific command for that.

kevinhwang91:
Thanks for the tip.

PR #3760, opened 2022-04-12
chore(README): fix show_documentation

afreakk:
&keywordprg expands to `:Man`.  
If prepended by `!`, causes:  
`bash: line 1: :Man: command not found`

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ function! s:show_documentation()
   elseif (coc#rpc#ready())
     call CocActionAsync('doHover')
   else
-    execute '!' . &keywordprg . " " . expand('<cword>')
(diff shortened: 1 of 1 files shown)


afreakk:
I use a `show_documentation` like this:
```
function! s:show_documentation()
  if (index(['vim','help'], &filetype) >= 0)
    execute 'h '.expand('<cword>')
  elseif (coc#rpc#ready())
    if CocAction('doHover') == 0
      execute  &keywordprg . " " . expand('<cword>')
    endif
  else
    execute &keywordprg . " " . expand('<cword>')
  endif
endfunction
```
So it falls back to manpage when hover is not found.  
But that might be too much for the `README.md`

PR #3764, opened 2022-04-16
chore(ci): run test with nvim 0.6.1/0.7.0

fannheyward:


chemzqm:
We should keep neovim 0.6.1 for sometime since neovim 0.7.0 just released.

fannheyward:
cc @chemzqm

PR #3777, opened 2022-04-22
feat(menu): add more mappings to confirm selection

xiyaowong:
`o` `<C-y>` `<C-o>` are convenient


chemzqm:
You can use `dialog.confirmKey` configuration.

xiyaowong:
Oops, didn't notice that option

PR #3783, opened 2022-04-26
fix(diagnostic): convert to byte index

kevinhwang91:


chemzqm:
Merged with fixed on 32cad713

kevinhwang91:
```typescript
            const max = diagnostics.reduce((p, c) => {
              return Math.max(c.range.end.line, p)
            }, 0)
            lines = await readFileLines(u.fsPath, 0, max)
```

太细了，佩服

PR #3789, opened 2022-04-28
fix(list): support ListTask for listLoadItems action

kevinhwang91:
Motivation:

I want to port result from `CocAction('listLoadItems', 'typos')` to quickfix and found that the source is `ListTask`.

If don't care the sink handler is in coc-extension or in Neo/vim, adding API for coc-extensions is better. The extension can get the ListTask and handle data by itself.  It's related to https://github.com/neoclide/coc.nvim/pull/3752 .


kevinhwang91:
我这个pr没改动api。目前是stream转promise再await，neovim或者vim那一侧异步就好。

PR #3819, opened 2022-05-14
feat(language-client): LSP 3.17

fannheyward:
LSP 3.17 highlighting features:

- [x] diagnostic pull model
- [x] inlay hints
- [x] type hierarchies
- [x] inline values
- [x] workspace symbol resolve request
- [x] completionList.itemDefaults, needs more tests, didn't find any LS supports this by now
- [ ] ~notebook documents~

**language-client API only**. `inlayHints` + https://github.com/neoclide/coc.nvim/commit/3a26df0abc7d432cca7a0ee6b8da3fc1ea5cae42 + rust-analyzer, works as expected.

@chemzqm for diagnostic pull model, `vscode-languageserver-node` mixed some VSCode tab functions, I've removed these, use `workspace.textDocuments/window.activeTextEditor` instead, and I think we should use `src/diagnostic` to handle the requests, please take a review.


chemzqm:
LGTM, but we need to finish features for 3.16 and publish new tag first.

There's only one feature remains:

- Annotations and transaction support of `WorkspaceEdit`

chemzqm:
More work needed to complete the support for LSP 3.17

- Create handler for new features.
- Update typings file.

fannheyward:
> Create handler for new features.

Yes, this PR is language-client API only, will add handler later.

chemzqm:
@fannheyward I will make some refactor to language-client module.

oblitum:
I believe this merge is the cause of many duplicated information in inlay hints? Is there an option to avoid this?

oblitum:
Hi @fannheyward, the screenshot above is not clangd, this time it's rust-analyzer, which didn't do that before.

oblitum:
It's being rather confusing to mentally parse what some of these inlay hints actually mean.

oblitum:
FWIW, the codebase on the screenshot is https://github.com/meli/meli.

fannheyward:
@oblitum it's the same reason for the duplicated hints, you need to disable from coc-rust-analyzer by now from `"rust-analyzer.inlayHints.enable": false`.

PR #3855, opened 2022-05-28
fix(jest): clean up tmp after exit

kevinhwang91:


jest.js
```diff
@@ -2,14 +2,18 @@ const path = require('path')
...
 const fs = require('fs')
 
-process.on('uncaughtException', function (err) {
+process.on('uncaughtException', err => {
   let msg = 'Uncaught exception: ' + err.stack
   console.error(msg)
 })
 
+process.on('exit', () => {
+  fs.rmdirSync(process.env.TMPDIR, { recursive: true, force: true })
```
fannheyward, line 11: This removes the whole TMPDIR, we should only remove `coc-test` that we created.

fannheyward (changes requested):

jest.js
kevinhwang91: But on the line 21 have changed the ENV

fannheyward (approved):

fannheyward:
Tested in my env, all tests passed but `error Command failed with exit code 1`.

kevinhwang91:
Like your nodejs version is below 14. I must AFK now.

fannheyward:
I'm using node 18.2.0

PR #3860, opened 2022-05-31
chore(inlayhint): use combine hl_mode

xiyaowong:
image


chemzqm:
Don't know if we need combine for inlay hint highlights @fannheyward

fannheyward:
> Don't know if we need combine for inlay hint highlights 

Sorry I totally missed this message. `combine` mode looks better with the background.

PR #3862, opened 2022-05-31
[WIP] Custom completion menu

chemzqm:
**This PR breaks current completion work flow**, the built-in completion of vim will be replaced with custom float window/popup.

Will be release with version `0.0.82`.

https://user-images.githubusercontent.com/251450/172049025-e49b65c0-a6f3-454f-b696-531a1abd4b82.mov

## Break changes

- Minimal version changed to nvim >= 0.4.0 and vim >= 8.1.1719.
- Removed configurations:  `suggest.enablePreview` `suggest.keepCompleteopt` `suggest.disableKind` `suggest.disableMenu` `suggest.floatEnable` `suggest.disableMenuShortcut`.
- `"suggest.noselect"` changed to `false` by default, like VSCode.
- `"suggest.enablePreselect"` changed to `true` by default, like VSCode.
-  Add `"suggest.formatItems"` for customize layout in popupmenu.
- Need to remap keys to manage completion list.

Default key-mappings are created to make the completion works without configuration.
```vim
if empty(mapcheck("\<C-n>", 'i'))
  inoremap <silent><expr> <C-n> coc#pum#visible() ? coc#pum#next(1) : "\<C-n>"
endif
if empty(mapcheck("\<C-p>", 'i'))
  inoremap <silent><expr> <C-p> coc#pum#visible() ? coc#pum#prev(1) : "\<C-p>"
endif
if empty(mapcheck("\<down>", 'i'))
  inoremap <silent><expr> <down> coc#pum#visible() ? coc#pum#next(0) : "\<down>"
endif
if empty(mapcheck("\<up>", 'i'))
  inoremap <silent><expr> <up> coc#pum#visible() ? coc#pum#prev(0) : "\<up>"
endif
if empty(mapcheck("\<C-e>", 'i'))
  inoremap <silent><expr> <C-e> coc#pum#visible() ? coc#pum#cancel() : "\<C-e>"
endif
if empty(mapcheck("\<C-y>", 'i'))
  inoremap <silent><expr> <C-y> coc#pum#visible() ? coc#pum#confirm() : "\<C-y>"
endif
if empty(mapcheck("\<PageDown>", 'i'))
  inoremap <silent><expr> <PageDown> coc#pum#visible() ? coc#pum#scroll(1) : "\<PageDown>"
endif
if empty(mapcheck("\<PageUp>", 'i'))
  inoremap <silent><expr> <PageUp> coc#pum#visible() ? coc#pum#scroll(0) : "\<PageUp>"
endif
```

## Example vimrc:

```vim
  inoremap <silent><expr> <CR> coc#pum#visible() ? coc#pum#confirm() : "\<C-g>u\<CR>\<c-r>=coc#on_enter()\<CR>"
  inoremap <silent><expr> <C-x><C-z> coc#pum#visible() ? coc#pum#stop() : "\<C-x>\<C-z>"
  " remap for complete to use tab and <cr>
  inoremap <silent><expr> <TAB>
        \ coc#pum#visible() ? coc#pum#next(1):
        \ <SID>check_back_space() ? "\<Tab>" :
        \ coc#refresh()
  inoremap <expr><S-TAB> coc#pum#visible() ? coc#pum#prev(1) : "\<C-h>"
  inoremap <silent><expr> <c-space> coc#refresh()

  hi CocSearch ctermfg=12 guifg=#18A3FF
  hi CocMenuSel ctermbg=109 guibg=#13354A
```
Checkout `:h coc-completion` for details.


oblitum:
I really like preselect feature. You state that `suggest.enablePreselect` configuration is removed, but the screenshot above seems like having preselect by default?

oblitum:
Nice, you should have this already planned, but I'd like to say would be nice to finally have borders for completion :)

oblitum:
Just pulled the branch and the borders are showing up out of the box fine (I have yet to remove round corner for completion). I notice though, that with border, there's a curious behavior, sometimes docs display quite far from completion, and a few time they're closer, which looks better than too far apart has happens most times it seems:

gif-2022-05-31-153010

oblitum:
Maybe that's side-effect of presence of scrollbar, in gif they're closer when there's no scrollbar.

oblitum:
I'm used to be able to scroll docs float with keyboard, it stopped working in this PR.

My config is:

```vim
inoremap <silent><expr> <c-j> coc#float#has_scroll() ? "\<c-r>=coc#float#scroll(1, 1)\<cr>" : "\<c-j>"
inoremap <silent><expr> <c-k> coc#float#has_scroll() ? "\<c-r>=coc#float#scroll(0, 1)\<cr>" : "\<c-k>"
```
On <kbd>\<c-j\></kbd> I'm getting:

```
autoload/coc/float.vim|657 error 121| function! coc#float#scroll(forward, ...): Undefined variable: id
autoload/coc/float.vim|657 error 116| function! coc#float#scroll(forward, ...): Invalid arguments for function getwinvar
autoload/coc/float.vim|657 error 121| function! coc#float#scroll(forward, ...): Undefined variable: id
autoload/coc/float.vim|657 error 116| function! coc#float#scroll(forward, ...): Invalid arguments for function getwinvar
```

chemzqm:
> it's the additional spacing before [LS]. It's not consistent, so things stop being aligned. It's not always before [LS], see this one on Rust (without dots):

I can't reproduce, what icon text you're using?

chemzqm:
Should be issue with your `"suggest.snippetIndicator"`, try use another one.

oblitum:
I changed `suggest.snippetIndicator` to a normal ascii char, and indeed, that has reduced the problem, but, it has not vanished, there's still cases left due to the unicode glyphs I pick for the "kind" column. If I make everything ascii chars, then the problem is gone, so, this indeed points to [incompatible unicode width calculations](https://github.com/tmux/tmux/issues/1057#issuecomment-358402953.

chemzqm:
@oblitum Set `"suggest.ambiguousIsNarrow": false` should fix your issue.

oblitum:
@chemzqm 

`"suggest.ambiguousIsNarrow": false`

img-2022-06-02-075340

`"suggest.ambiguousIsNarrow": true`

img-2022-06-02-075629

Thanks for the attempt. Sorry but it didn't help much in my test situation, I think your changed made it better by default (`"suggest.ambiguousIsNarrow": true`), where at least the left of 2nd column is always left aligned, but aligned is still missed in the 3rd column. When it's `false` it's actually worse.

I'm currently just [avoiding explicit emoji representation](https://codeberg.org/oblitum/dotfiles/commit/b07afdb6a715558b00063031909f8d8e9b4d55d5) to evade the issue, it doesn't have anymore when I do that. I actually had to revert those changes locally to test the new `ambiguousIsNarrow` option.

oblitum:
Completion with border at first column of window causes the text and scrollbar to shift left, overwriting the border on the left, and leaving space after scrollbar on the right.

img-2022-06-02-080831

chemzqm:
> `"suggest.ambiguousIsNarrow": true`
> 
> img-2022-06-02-075629

Try only use single width icon for `"suggest.completionItemKindLabels"`

kevinhwang91:
How can I customize the menu? Have a Wiki or doc for help?

chemzqm:
See `:h coc-completion`

oblitum:
This seems to happen with certain frequency, specially at middle of screen:

gif-2022-06-04-220005

The file I'm editing above is [this](https://codeberg.org/oblitum/dotfiles/src/branch/%EF%8C%83/.config/nvim/init.lua), with coc-sumneko-lua.

The experience isn't great as I get confused about where to look at (besides looking like fireworks 😅). It happens because, when there's few completion items, pum shows below, if there's too much, then it goes down or up, depending which has more room to contain more completion results I guess, I have yet to check configuration for pum result limit, to try reduce the problem.

Besides this, I think it would be a good option for Coc if it provided a flag to make the pum behave rigidly, like, "always bellow", no matter if there's just two lines visible bellow the cursor, because, even though it is a nice feature to have smart pum that goes down or up, and even if that's implemented more correctly and stable, there's always some kind of users that like stuff always where it expects, and if there's no room, they prefer to scroll to provide room than to see it showing up above, for example.

oblitum:
Duplicated doc added [here](https://github.com/neoclide/coc.nvim/pull/3862/commits/0250de9bbf520b5476d4840caf862dcad95ca33b#diff-26c6f4c36a5b8d2caa2551a2461799e9cd08f3e8ffbb0cefb5f0a7838791debdR203).

autoload/coc/pum.vim
```diff
@@ -0,0 +1,458 @@
...
+    if strchars(word) > strchars(input) && strcharpart(word, 0, strchars(input)) ==# input
+      let insert = strcharpart(word, strchars(input))
+    endif
+    call nvim_buf_clear_namespace(bufnr, s:virtual_text_ns, 0, -1)
+    if !empty(insert)
+      let opts = {
+          \ 'virt_text': [[insert, 'CocPumVirtualText']],
+          \ 'virt_text_pos': 'overlay',
+          \ 'virt_text_win_col': col('.') - 1,
+          \ }
```
xiyaowong, line 448: combine mode would be better

oblitum:
This PR's branch has an issue where overall diagnostics information (in buffer, signcolumn, status) gets stuck after first diagnostic refresh:

gif-2022-06-05-142913

chemzqm:
It's fixed on master.

autoload/coc/pum.vim
chemzqm: Ok

src/completion/complete.ts
```diff
@@ -118,20 +109,38 @@ export default class Complete {
...
+    if (variables.blacklist?.length) {
+      if (variables.blacklist.includes(this.option.input)) {
+        logger.warn('suggest cancelled by b:coc_suggest_blacklist')
+        return true
+      }
+    }
     let { triggerCompletionWait, localityBonus } = this.config
     await wait(Math.min(triggerCompletionWait ?? 0, 50))
-    if (token.isCancellationRequested) return true
+    if (token.isCancellationRequested) return
```
fannheyward, line 137: Useless `token.isCancellationRequested` checking because it has been checked in L115.

chemzqm: Need to check again, because there's await between.

plugin/coc.vim
```diff
@@ -640,7 +658,7 @@ nnoremap <silent> <Plug>(coc-float-hide)            :<C-u>call       coc#float#c
 nnoremap <silent> <Plug>(coc-float-jump)            :<c-u>call       coc#float#jump()<cr>
 nnoremap <silent> <Plug>(coc-command-repeat)        :<C-u>call       CocAction('repeatCommand')<CR>
 nnoremap <silent> <Plug>(coc-refactor)              :<C-u>call       CocActionAsync('refactor')<CR>
-inoremap <silent>                          <Plug>CocRefresh <C-r>=coc#_complete()<CR>
+inoremap <silent> <Plug>CocRefresh <C-r>=coc#_complete()<CR>
```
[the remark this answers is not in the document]
chemzqm: It was used for feedkey to trigger completion, not needed any more.

oblitum:
> don't know why your languge server added a dot.

FYI,

```
❯ clangd --help | rg 'decorators'
  --header-insertion-decorators   - Prepend a circular dot or space before the completion label, depending on whether an include line will be inserted or not
```

fannheyward (approved):

SteveCharleston:
This looks very nice so far, however one things breaks my workflow completely. With the normal pum if I type something and the suggestions start to appear, if I hit `<C-n>` then the first suggestion will be inserted and I can continue typing more text. With this new menu, the first item is pre-selected but of course nothing is inserted. If I now hit `<C-n>` the second item is selected and inserted, so I now have to hit `<C-p>` to get back to the first suggestion and have it inserted.

Is there an easy map or configuration option to select and insert the first item, if I hit `<C-n>` the first time after the menu appeared?

oblitum:
@SteveCharleston `<c-y>` is used for selection confirmation (has always been), although it expands snippets if item is snippet. Previously you could get to the text as you say, without triggering snippet expansion, that's the option that seems missing.

SteveCharleston:
Would be nice if either this option would be provided again or at least someone would know some workaround to have `<C-n>` select the first item on first press after the menu appeared. Since normally only this shortcut was needed to reach all suggestions, now it's either `<C-y>` for the first or `<C-n>` for later ones.
It might seem to not be much, but it really breaks the flow for me right now (and as said above, at least a workaround that someone might come up with would be helpful)

oblitum:
Another thing that I'm noticing, and I'm not sure it's new behavior or not, is that, after typing half a word, and navigating with `<c-n>`/`<c-p>`, even if I don't pick any item and just cancel completion with `<c-e>`, next time I type this half word, the pum will remember the last position/word I have stopped, it won't have first item selected. This doesn't happen when using arrows to browse, as their mapping don't insert text by default.

ratheesh:
`pum` of neovim has `spell` enabled. We may have to disable this.

This feature is very nice! Many thanks! The only nice to have feature missing on the visual side is `shadow text` that is supported in `nvim-cmp` and `coq_nvim`

lmburns:
In this commit I see that the colors of the popup-menu have changed. I can't seem to figure out which highlight group it is that is being linked to the unmatched portion of the text.

`CocPumSearch` = highlight matched input characters
`CocPumMenu` = highlight source (e.g., `[TS]`)
`CocMenuSel` = highlight background of current selection

Sorry if I am missing this somewhere

oblitum:
@lmburns `.borderhighlight` to `Normal` for everything.

chemzqm:
> e-selected but of course nothing is inserted. If I now hit `<C-n>` the second item is selected and inserted, so I now have to hit `<C-p>` to get back to the first suggestion and have it inserted.

Use `coc#pum#confirm()` or `coc#pum#insert()`

chemzqm:
@lmburns you can link higlight group `CocFloating` to `Normal`

chemzqm:
> next time I type this half word, the pum will remember the last position/item I have stopped

Fixed.

SteveCharleston:
> > e-selected but of course nothing is inserted. If I now hit `<C-n>` the second item is selected and inserted, so I now have to hit `<C-p>` to get back to the first suggestion and have it inserted.
> 
> Use `coc#pum#confirm()` or `coc#pum#insert()`

It's more about the workflow, previously it was possible to cycle through all suggestions and select them with `<C-n>`. Now it's either another key that executes `coc#pum#confirm()` for the first item or `<C-n>` for all other items. It would be nice to still have only `<C-n>` to reach and insert all suggestions, instead of having to differentiate between the first one and the other ones.

chemzqm:
> It would be nice to still have only <C-n> to reach and insert all suggestions, instead of having to differentiate between the first one and the other ones.

It's not an issue with `<C-n>`, but you want the behavior of `noselect`, so you have to go through the list by `<C-n>`, is that right?

We can add `noselect` back, but we can't give better suggestion on that case.

SteveCharleston:
> It's not an issue with `<C-n>`, but you want the behavior of `noselect`, so you have to go through the list by `<C-n>`, is that right?
> 
> We can add `noselect` back, but we can't give better suggestion on that case.

Yes you are right, it's `noselect` I wanted back in the end. I checked out your latest changes and enabled `suggest.noselect` and it's now working perfect for me again.

Thank you very much for taking care!

oblitum:
@chemzqm would it be nice to have an option to have client-only preselection, to ignore the server's preselect item when it provide support for that? The issue is that, with current implementation, if server has no support for that, client just preselect first suggestion, which is okay, but if some server at random provides support for preselect, what happens is that now, some random item in the completion list, may or may not be preselected, if server doesn't send anything, then it's first item by client, but if server sends something, it may be nth item in the completion list. This is a rather fuzzy behavior in my opinion, some people may enjoy the server trying to help further when it can, but I personally don't enjoy that that brings such unexpected eye jumping, worse even, at random times.

kevinhwang91:
Found an issue about `pumwidth`. 
If its value exceeds the pum width, the kind will not be right aligned.

image

chemzqm:
> Found an issue about `pumwidth`. If its value exceeds the pum items width, the kind will not be right aligned.

It's expected.

derekstavis:
Tested this branch and it works!

I noticed that when the popup opens too close to the edge of the screen, it is compressed. It would be nice to shift the popup to fit instead.

image

statiolake:
Thank you, it works well!

I found that dot-repeat no longer works even if completed item is not a snippet. [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) has limited support of it by simulating key input with `feedkeys()` instead of setting text through API. (Or we can use [vim-repeat](https://github.com/tpope/vim-repeat) ...? I'm not sure.)

chemzqm:
@statiolake `feedkeys` could cause more issue, vim's native completion not work with repeat when using `<C-n>` and `<C-p>` as well.

statiolake:
@chemzqm Thank you for your reply. Definitely `feedkeys()` introduces a lot of complexity...

Note that Neovim's native completion (and current coc.nvim) works with dot repeat.

Vanilla Neovim:

noplugin

With current coc.nvim:

pum

With this branch:

nopum

In the screencasts, I did roughly the following things:

1. Enter insert mode with `i`, entering prefix and select `world` from completion.
2. Leave insert mode and repeat several times with `.`. Vanilla and current coc.nvim inserts `world` at the cursor position, but custom pum inserts only `wor` (actually typed characters instead of completed one).
3. Replace current word with word by `ciw`. Within insert mode, enter `wo` and select completion `word`. Observe that vanilla and current coc.nvim correctly repeats word replace at another place, but custom pum only replaces to `wo` (typed characters, not completed one).

chemzqm:
@derekstavis I've made a fix.

chemzqm:
> @chemzqm Thank you for your reply. Definitely `feedkeys()` introduces a lot of complexity...

It's not complexity, it's issues with other vim plugins.

> Note that Neovim's native completion (and current coc.nvim) works with dot repeat.

Dot repeat won't work when it's not strict match.

statiolake:
@chemzqm I understand. Do you think it's possible to support the built-in behavior with custom pum? If not, do you have any plans to keep the built-in pum along with the custom one?

chemzqm:
> @chemzqm I understand. Do you think it's possible to support the built-in behavior with custom pum? 

It could support redo by use `complete` function, but there still some issues when using `complete` to insert text.
Maybe you can find out some way to insert the text with repeat support (without feedkeys) or you can send feature request to vim

> do you have any plans to keep the built-in pum along with the custom one?

No, you will have to use old version of coc.nvim for builtin pum.

chemzqm:
@statiolake I've made the change repeatable by use `complete`

statiolake:
@chemzqm Awesome! It works nicely. Thank you!

leoentersthevoid:
how does one disable the pop up saying "_coc.nvim switched to custom popup menu from 0.0.82..._"?

weirdsmiley:
I am using the above mentioned ex commands:
```vim
...
inoremap <silent><expr> <TAB>
        \ coc#pum#visible() ? coc#pum#next(1):
        \ <SID>check_back_space() ? "\<Tab>" :
        \ coc#refresh()
...
```

But this is not mapping Tab. Though, all other bindings mentioned (CR, S-Tab, C-Space..) are being mapped.

I have tried putting an `iunmap <Tab>` before these commands still no benefit. The only way I could make that mapping work was by sourcing the specific file (which contained those above commands) again `:source <file>`. Then only, it can map `<Tab>`.

I find this way tedious. Is there any fix/way around this?

chemzqm:
Use verbose command like `:verbose imap <tab>` to checkout  where it remapped.

weirdsmiley:
I did that, and it showed me a different plugin where it was being mapped. But the catch is, those mappings are being done prior to these coc mappings. AFAIK these should overwrite them, but seemingly they don't.

chemzqm:
> those mappings are being done prior to these coc mappings.

They're not.

JaoSchmidt:
There is a weird behaviour where the popup pre-selects the first item without writing it. Instead writing the second after the user hits <Tab>.
The only way to fix seems to apply  ```suggest.noselect": true``` on coc-settings.json.
Not sure if its intended

chemzqm:
It is intended to make suggest.selection could work

brandonpittman:
After this update, my pop-up menu appears cut-off…

CleanShot 2022-08-17 at 10 45 48@2x

ahnafalnafis:
@chemzqm, how to make matched character highlighted and also colorize the kindLabels?
image

chemzqm:
https://github.com/neoclide/coc.nvim/blob/a2df971b0656c6dbf4172aab417420f6a7bbcfbd/doc/coc.txt#L402

MathiasKandelborg:
Stop giving me popups, what the hell guys

alexbr:
I'll admit my vimscript isn't great, so I may be missing something here.

This seems to break plugins that depend on `pumvisible()` and don't, and shouldn't, know anything about `coc#pum#visible()`. 

For instance, I use vim-autoclose, which [remaps navigation keys in insert mode](https://github.com/Townk/vim-autoclose/blob/master/plugin/AutoClose.vim#L471) to an expression that checks `pumvisible()` to see if it should just return the default mapping, eg. `<up>`. With this update, when in the coc autocomplete menu and I press up and down, I get my coc mapping of up and down, `coc#pum#visible() ? coc#pum#next(0) : "\<up>"` echoed to the editor instead. 

Output of `:verbose imap <up>`:
```
i  <Up>        *@pumvisible() ? '<Up>' : '<C-R>=<SNR>83_FlushBuffer()<CR>coc#pum#visible() ? coc#pum#prev(0) : "\<Up>"<CR>' 
        Last set from ~/.vim/plugged/vim-autoclose/plugin/AutoClose.vim line 471
i  <Up>        * coc#pum#visible() ? coc#pum#prev(0) : "\<Up>"
        Last set from ~/.config/nvim/init.vim line 711
```

Granted, that's a bit of a wonky remap on that plugin's part, but the point remains the same, the expectation that `pumvisible()` should work as expected is now broken.

So, I guess what I'm asking is, will this update fix the native `pumvisible()` so that plugins that care about the popup menu being visible still work?

chemzqm:
@alexbr You have to create the key-mappings you need for that case, or use coc.nvim 0.0.81

Krzmbrzl:
Is there a way to change the popup's color?

bag-man:
I'm having a hard time translating my old config to use this new menu, here is my old config, working with neovim@0.8.0 and  coc@ 0.0.81. 

```
  " Autocomplete navigation
  inoremap <expr> J ((pumvisible())?("\<C-n>"):("J"))
  inoremap <expr> K ((pumvisible())?("\<C-p>"):("K"))
  
  "" use <tab> for trigger completion and navigate to the next complete item
  function! s:check_back_space() abort
    let col = col('.') - 1
    return !col || getline('.')[col - 1]  =~ '\s'
  endfunction
  inoremap <silent><expr> <Tab>
        \ pumvisible() ? "\<C-n>" :
        \ <SID>check_back_space() ? "\<Tab>" :
        \ coc#refresh()
```

How can I recreate these bindings?

Thanks!

fannheyward:
Check example on README or `:h coc-completion-example`.

chemzqm:
Replace `pumvisible()` with`coc#pum#visible()`.
Replace `"\<C-n>"` with `coc#pum#next(1)`.
Replace `"\<C-p>"` with `coc#pum#prev(1)`.

skywb:
Is can't complete from symbol [A] ?
image

PR #3865, opened 2022-05-31
Fixes: #3864 UnhandledRejection: Cannot read property 'length' of null

ronakg:
This PR fixes the issue described in https://github.com/neoclide/coc.nvim/discussions/3864

The language server may not provide the semantic tokens always. So this fix adds protection for such cases.


chemzqm:
I've made a fix, check the lint error.

PR #4024, opened 2022-08-09
Remove Search highlight in floats

creasty:
## Why

Remove `Search` highlight added by `/` command in floats.

See the example:
https://github.com/neoclide/coc.nvim/discussions/4016#discussioncomment-3345247

## What

Adds `Search:` to `&winhl`.
Note that it's intentional that the second argument is empty, which effectively removes the definition of the specified highlight.


chemzqm:
@all-contributors please add @creasty for code.

PR #4025, opened 2022-08-09
Merge &winhl rather than just replace it

creasty:
## Why

When a float window is reused, it forcibly resets `&winhl` and never fires a CocOpenFloat event.

https://github.com/neoclide/coc.nvim/discussions/4016#discussioncomment-3345247

## What

Add the default value by merging it into the current `&winhl`.


chemzqm:
Improve the implementation on 578d2e34

Issue #4039, opened 2022-08-10
seems remain last selected item while no input


springhack:
## Result from CocInfo

```
vim version: VIM - Vi IMproved 8.2 8023995
node version: v18.7.0
coc.nvim version: 0.0.82-578d2e34 2022-08-10 02:05:48 +0800
coc.nvim directory: /root/.vim/plugged/coc.nvim
term: dumb
platform: linux
```

## Describe the bug

When trigger without input, select any item then select none, it remain the last selected item.

## Reproduce the bug

- Create file `mini.vim` with：

```vim
set nocompatible
set runtimepath^=~/.vim/plugged/coc.nvim
filetype plugin indent on
syntax on
set hidden

func! s:check_back_space() abort
  let col = col('.') - 1
  return !col || getline('.')[col - 1]  =~# '\s'
endfun

inoremap <silent><expr><down>  coc#pum#visible() ? coc#pum#next(1) : "\<down>"
inoremap <silent><expr><up>    coc#pum#visible() ? coc#pum#prev(1) : "\<up>"
inoremap <silent><expr><tab>   coc#pum#visible() ? coc#pum#next(1) : <SID>check_back_space() ? "\<tab>" : coc#refresh()
inoremap <silent><expr><s-tab> coc#pum#visible() ? coc#pum#prev(1) : "\<s-tab>"
```

- Start (neo)vim with command: `vim -u mini.vim`

- Install coc-tsserver and typescript npm package, some code like in screenshot(gif)

## Screenshots (optional)

![https://i.imgur.com/h0r40cs.gif](https://i.imgur.com/h0r40cs.gif)

PR #4042, opened 2022-08-11
fix(completion): reset input word even it is empty string

springhack:
fix this #4039


chemzqm:
@all-contributors please add @springhack for bug.

PR #4064, opened 2022-08-17
feature(typings): expose floatConfig and applyFloatConfig

ytian81:
This PR exposes floatConfig interface and applyFloatConfig method so that coc-git can use hoverFloatConfig.

https://github.com/neoclide/coc-git/pull/199#issuecomment-1217285961

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -4652,6 +4652,20 @@ declare module 'coc.nvim' {
     readonly options: TextEditorOptions
   }
 
+  export interface FloatConfig {
+    border?: boolean
+    rounded?: boolean
+    highlight?: string
+    title?: string
+    borderhighlight?: string
+    close?: boolean
+    maxHeight?: number
+    maxWidth?: number
+    winblend?: number
+    focusable?: boolean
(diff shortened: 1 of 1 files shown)


chemzqm:
We should use a default `floatConfig` configuration for `floatFactory` instead, and you can copy that function if needed.

ytian81:
Wondering which function are you suggesting? The `floatFactory` show function only takes `FloatWinConfig`


https://github.com/neoclide/coc.nvim/blob/d9223af57af7c72cba0fe48c39ba56b1c84e292b/src/model/floatFactory.ts#L136-L150

chemzqm:
You don't need `applyFloatConfig`, `FloatWinConfig` is enought

ytian81:
My local changes to coc-git is 

image


For sure I could transform `hover.floatConfig` to `FloatWinConfig` without `applyFloatConfig` if that's what you mean. Basically duplicating the logic of `applyFloatConfig` instead of exporting it in coc.vim

chemzqm:
Yes, copy the code instead.

PR #4091, opened 2022-08-24
Fix formatting usage in example

kristofer-lp8f:
In the example given in the README, formatting in normal mode works only with a movement. This PR changes to formatting selected text (which was probably the intention), and formatting in normal mode formats the entire file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,8 @@ nmap <leader>rn <Plug>(coc-rename)
 
 " Formatting selected code.
 xmap <leader>f  <Plug>(coc-format-selected)
-nmap <leader>f  <Plug>(coc-format-selected)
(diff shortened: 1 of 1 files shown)


chemzqm:
It's `Formatting selected code`, command is better for format current file.

kristofer-lp8f:
Why is `coc-format-selected` better for formatting the entire file?

Example:
I'm in the middle of a file.
- This PR: `<leader>f`
- Current: `gg<leader>fG`, and I lose my position in the file

chemzqm:
It's an example for `format selectedrange`

kristofer-lp8f:
Then the example is still incorrect. The mapping should be `vmap`, not `nmap`, if you want to target the _selected_ range

PR #4132, opened 2022-09-01
feat(diagnostic): Support "diagnostic.virtualTextFormat" to define virtual text format

qiqiboy:
### Two changes

1. Add `diagnostic.virtualTextFormat` (maybe `virtualTextPrefix` can be deprecated?)
2. Fix virtual text diagnostics shown order `arr.push(...)` --> `arr.unshift(...)`. This is because the diagnostics have been sorted by `severity`, but the `for ... in` loop is reverved. This change ensures that high-severity diagnostic message shown first

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -782,6 +782,11 @@
       "description": "The prefix added virtual text diagnostics",
       "default": " "
     },
+    "diagnostic.virtualTextFormat": {
+      "type": "string",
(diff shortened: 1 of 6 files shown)


chemzqm:
There're some test failures.

> maybe virtualTextPrefix can be deprecated?

Not needed.

chemzqm (changes requested):
Please fix tests

PR #4134, opened 2022-09-02
docs(help): fix CocInlayHint setting sentences.

seiyeah78:
ref: 38a95c8c0


chemzqm:
It's not link to SignColumn as well, will fix that.

PR #4143, opened 2022-09-03
fix(plugin): should initialize highlight groups properly

kevinhwang91:
May miss `VimEnter` if lazy load coc.nvim


chemzqm:
Thanks, fixed by 71602882

PR #4208, opened 2022-09-17
Keep empty lines in markdown

HampusMat:
How users choose to format their documentation should be of no concern to this plugin.

You can probably agree on how hard it is to read the following
bad-hover

The original markdown is
```
Creates a binding of type `Interface` to type `Implementation` inside of the
associated [`DIContainer`](https://docs.rs/syrette/0.3.0/syrette/di_container/struct.DIContainer.html).

The scope of the binding is transient. But that can be changed by using the
returned [`BindingScopeConfigurator`](https://docs.rs/syrette/0.3.0/syrette/di_container/struct.BindingScopeConfigurator.html)

# Errors

Will return Err if the associated [`DIContainer`](https://docs.rs/syrette/0.3.0/syrette/di_container/struct.DIContainer.html) already have a binding for
the interface.']
```

The below is with the commit in this pull request
good-hover

### DIFF
--- a/src/__tests__/markdown/index.test.ts
+++ b/src/__tests__/markdown/index.test.ts
@@ -89,14 +89,14 @@ example:
 \`\`\`
     `
     let res = parseMarkdown(content, {})
-    expect(res.lines).toEqual(['example:', '<div>code</div>'])
-    expect(res.codes).toEqual([{ filetype: 'html', startLine: 1, endLine: 2 }])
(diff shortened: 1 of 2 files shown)


chemzqm:
I think detect for next line is need, if it's a code block or `<hr>`, we'd better remove the blank lines.

chemzqm:
Should be fixed by 043e06c2, empty lines are kept when necessary.

PR #4221, opened 2022-09-22
perf(highlight.lua): add lua highlight to reduce blocking

kevinhwang91:
For now,  coc.nvim use Vimscript to drive highlighting, but Vimscript is slow enough. After opening https://github.com/tree-sitter/tree-sitter-javascript/blob/936d976a782e75395d9b1c8c7c7bf4ba6fe0d86b/src/parser.c#L70311, coc.nvim makes neovim lagging.

The bottleneck is in https://github.com/neoclide/coc.nvim/blob/488d6ae762f905f58a362294bab0af2150d0c399/autoload/coc/highlight.vim#L737-L743

TBH, I hope coc.nvim can hug the Lua.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,3 +20,15 @@ indent_style = space
 indent_size = 2
 trim_trailing_whitespace = true
 max_line_length = 120
+
+
+[*.lua]
+indent_size = 2
+max_line_length = 120
+align_call_args = only_not_exist_cross_row_expression
+align_table_field_to_first_field = true
+local_assign_continuation_align_to_first_expression = true
+keep_one_space_between_table_and_bracket = false
+quote_style = single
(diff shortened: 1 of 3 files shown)


kevinhwang91:
I'm not sure whether coc.nvim has a plan to use a Lua highlight module to drive `extmarks`. I think it's meaningful because the perf and the APIs about `extmarks` are messy enough so using Vimscript mix with `extmarks` to be compatible with Vim is painful to a contributor like me.

chemzqm:
> I'm not sure whether coc.nvim has a plan to use a Lua highlight module to drive `extmarks`. 

We will use lua as long as we could have some performance improvement.
And we will use vim9 for some function for better performance.

PR #4237, opened 2022-09-26
Improve readibility of readme.md

CollieIsCute:
- use `vim.keymap.set()` to replace `vim.api.nvim_set_keymap()`
	- thanks to [this commit](https://github.com/neovim/neovim/commit/6d41f65aa45f10a93ad476db01413abaac21f27d), `vim.keymap.set` is higher level than `vim.api.nvim_set_keymap`.
	- One obvious benefit is that we can direct call lua function in `vim.keymap.set`
- adjust some position of code

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,37 +262,30 @@ vim.opt.updatetime = 300
 -- diagnostics appear/become resolved.
 vim.opt.signcolumn = "yes"
 
-
+local keyset = vim.keymap.set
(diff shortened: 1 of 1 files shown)


25077667 (commented):
Seems okay.

chemzqm:
@all-contributors please add @CollieIsCute for doc.

PR #4291, opened 2022-10-15
docs(snippet): complete usage of two functions

timsu92:


chemzqm:
Thanks, but those functions doesn't requires ``coc-snippets`.

PR #4324, opened 2022-10-26
fix(diagnostic): dirty version should be compared with last version

kevinhwang91:
Fix the issue like the below video:
https://user-images.githubusercontent.com/17562139/197919963-86c29ccd-dab0-4638-8476-1e15c7aa916a.mp4

Text may be changed while refreshing, should use a new var to compare with the dirty version.


chemzqm:
The gopls is wired as it may not send the same diagnostics after content change, I've fixed the issue.

PR #4325, opened 2022-10-26
Support configuration registry

chemzqm:
Store configuration schemas.

Since coc.nvim have only supported 3 scoped, some conventions are done for configuration from property scope in package.json:

- `application` or `window` => `window`
- `resource` or `machine-overridable` => `resource`
- `language-overridable` => `language-overridable`
-  `undefined` or other cases  => `resource`


NoahTheDuke:
Hey @chemzqm, this seems to have broken things for coc-clojure. I'm not sure exactly what's happening in this PR so if you would be so kind, give me a rough description so I can attempt to fix things on my end?

chemzqm:
@NoahTheDuke default value are created for configuration properties without default value (same as VSCode), the clojure server crash with `  "clojure.initialization-options.cache-path": ""`, set the value the `null` or filter out bad configuration in coc-clojure should fix the issue.

PR #4355, opened 2022-11-11
Fix coc#snippet#jump to factor in coc#pum#visible

creasty:
This PR addresses a bug where `"suggest.preferCompleteThanJumpPlaceholder": true` is ineffective with the custom PUM.


creasty:
Not sure what's going on here...

```ts
      await session.start('${1|one,two,three|}', defaultRange)
      let line = await nvim.line
      expect(line).toBe('one')
      await helper.waitPopup() //<<<<<<<<<<<<<
```

chemzqm:
Thanks, the CPU of CI could be too busy sometimes.

PR #4384, opened 2022-11-22
Update `CocSem*` and `CocSymbol*` Highlight Links

LebJe:
Link `CocSem*` and `CocSymbol*` to `@*` [because nvim-treesitter removed `TS*`](https://github.com/nvim-treesitter/nvim-treesitter/pull/3656).

Highlights like `@file`, `@package` `@array` don't exist, but color schemes can add highlights for them.


plugin/coc.vim
```diff
@@ -531,31 +531,31 @@ function! s:Highlight() abort
...
-        \ 'Regexp': ['TSStringRegex', 'String'],
-        \ 'Operator': ['TSOperator', 'Operator'],
-        \ 'Decorator': ['TSSymbol', 'Identifier'],
-        \ 'Deprecated': ['TSStrike', 'CocDeprecatedHighlight']
+        \ 'Namespace': ['@namespace', 'Include'],
+        \ 'Type': ['@type', 'Type'],
+        \ 'Class': ['@constructor', 'Special'],
+        \ 'Enum': ['@type', 'Type'],
+        \ 'Interface': ['@type', 'Type'],
+        \ 'Struct': ['@type', 'Identifier'],
```
oxalica, line 539:
```suggestion
        \ 'Struct': ['@structure', 'Identifier']
```

Also `Enum`, `Interface`, `EnumMember` and etc used to have different TS* hl names, while how they are merged into same `@type` or `@property`. Is this intended?

oxalica (changes requested):

plugin/coc.vim
LebJe: When colorschemes updated ([1](https://github.com/navarasu/onedark.nvim/pull/110/files), [2](https://github.com/ellisonleao/gruvbox.nvim/pull/163)), they did not add `Enum`, `Interface`, `structure`, or `EnumMember` highlights. Also, none of those highlights are mentioned in [nvim-treesitter’s CONTRIBUTING.md ](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#highlights).

oxalica: But unlike others, `@structure` is a thing shipped with neovim 0.8.1, which links to `Structure` by default.
See `:help treesitter-highlight`.

oxalica (approved):

fannheyward:
The `@*` hi groups only available in nvim 0.8+, this means change `TS*` to `@*` drops nvim 0.7 supports. Do we need to support both? cc @chemzqm 

I prefer to drop the support because:

1. users always use nvim latest release
2. `CocSem*` will fallback to `String` etc if treesitter not available

chemzqm:
> I prefer to drop the support because:
> 
> 1. users always use nvim latest release
> 2. `CocSem*` will fallback to `String` etc if treesitter not available

Agreed, don't think we need support old version TS highlights.

fannheyward:
cc @chemzqm, do you still have any questions on this change?

chemzqm:
@all-contributors please add @LebJe for code.

PR #4416, opened 2022-12-07
feat(list): avoid highlight colorcolumn in list window

skysky97:
We don't need to show colorcolumn in list window.


chemzqm:
@all-contributors please add @skysky97 for bug.

PR #4436, opened 2022-12-13
feat(extension): prompt to install recommend extensions

fannheyward:
A very minimal implementation.

`extensions.recommendations` can be set with a list of extensions' name, coc will prompt user to install or not.

How VSCode does this: https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

1. workspace only, not global
2. uses `.vscode/extensions.json` file to locate the settings, in workspace folder
3. `recommendations` for should be recommended, `unwantedRecommendations` for should not be recommended for users of this workspace

@chemzqm any suggestions? Do we need to use a separate file to save the settings?

### DIFF
--- a/src/extension/index.ts
+++ b/src/extension/index.ts
@@ -80,6 +80,7 @@ export class Extensions {
     await this.manager.activateExtensions()
     let names = this.states.filterGlobalExtensions(workspace.env.globalExtensions)
     void this.installExtensions(names)
+    await this.promptRecommendedExtensions()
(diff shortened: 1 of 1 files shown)


chemzqm:
> Do we need to use a separate file to save the settings?

I prefer not, use a section for settings.json should be enough.

Other stuff for this feature:

* Prompt only once for single workspace folder. 
* doc and test

PR #4450, opened 2022-12-22
fix(util): detect fieltype if none

fannheyward:
`deno lsp` returns definitions like `deno:/asset/lib.es5.d.ts`, coc-deno will register `provideTextDocumentContent` to fetch and return content, coc will set buf with content and fire `BufCreate` to `createDocument`.

In this case, `coc#util#get_bufoptions` buffer filetype is none, we should detect it.

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -90,7 +90,7 @@ endfunction
 function! coc#util#path_replace_patterns() abort
   if has('win32unix') && exists('g:coc_cygqwin_path_prefixes')
     echohl WarningMsg
-    echon 'g:coc_cygqwin_path_prefixes is deprecated, use g:coc_uri_prefix_replace_patterns instead' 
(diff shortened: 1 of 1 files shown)


fannheyward:
This change broken `Start server on document open`, have no idea about this...

chemzqm:
It's not a good place to detect filetype, I've made a fix.

fannheyward:
@chemzqm sorry for the delay, but that fix didn't work for this case.

Test steps:

1. coc.nvim + deno 1.29.1 + coc-deno
2. `echo 'const arr = new Uint8Array(100);' > main.ts`
3. `jumpDefinition` on `Uint8Array`, will list `lib.es5.d.ts` etc
4. jump to any file, no filetype detected

chemzqm:
Seems you're using something else to open the buffer instead of API from coc.nvim

PR #4455, opened 2022-12-24
doc: follow implementations to replace CocAction with CocActionAsync

jiz4oh:


chemzqm:
There're not tags for `CocActionAsync('diagnosticInfo')` etc.

PR #4488, opened 2023-01-30
fix(handler): use proper endline for inner function

### LINKED ISSUE #1847: Incorrect ranges from coc-funcobj-a/i
## Result from CocInfo

```
## versions

vim version: NVIM v0.4.3
node version: v13.13.0
coc.nvim version: 0.0.78-fc983ce6dd
term: rxvt-unicode-256color
platform: linux

## Output channel: clangd log

I[05:03:40.186] clangd version 10.0.0 
I[05:03:40.186] PID: 274389
I[05:03:40.186] Working directory: /tmp/coc-test
I[05:03:40.186] argv[0]: /usr/bin/clangd
I[05:03:40.186] Starting LSP over stdin/stdout
I[05:03:40.187] <-- initialize(0)
I[05:03:40.187] --> reply:initialize(0) 0 ms
I[05:03:40.190] <-- initialized
I[05:03:40.190] unhandled notification initialized
I[05:03:40.191] <-- textDocument/didOpen
I[05:03:40.191] Failed to find compilation database for /tmp/coc-test/main.cpp
I[05:03:40.191] --> textDocument/clangd.fileStatus
I[05:03:40.191] Updating file /tmp/coc-test/main.cpp with command clangd fallback
[/tmp/coc-test]
/usr/bin/clang /tmp/coc-test/main.cpp -fsyntax-only -resource-dir=/usr/lib/clang/10.0.0
I[05:03:40.201] --> textDocument/publishDiagnostics
I[05:03:40.201] --> textDocument/clangd.fileStatus
I[05:03:42.004] <-- textDocument/documentSymbol(1)
I[05:03:42.004] --> reply:textDocument/documentSymbol(1) 0 ms
I[05:03:42.004] --> textDocument/clangd.fileStatus
```

## Describe the bug

The text objects `coc-funcobj-a` and `coc-funcobj-i` give incorrect ranges for some language servers.
I observed this with `coc-clangd` and `coc-rls`.

For the `coc-funcobj-i` text object, this code just takes the range for the complete function and removes the first and last line, which doesn't work for function signatures spanning multiple lines:
https://github.com/neoclide/coc.nvim/blob/d6dd56bba8992aae3ade4034f1bc8d1148c3780f/src/handler/index.ts#L817-L822
It's probably not possible to do this more accurately without LSP support, but some documentation about it would be helpful. I was initially unsure whether this was an issue with coc or with clangd.

I have no idea why `coc-funcobj-a` doesn't work, though. Reading the code, it seems to just use the range reported by the server, but the end position is always one line too high.

## Reproduce the bug

- Create file `mini.vim` with:

  ```vim
  set nocompatible
  filetype plugin indent on
  syntax on
  set hidden
  xmap if <Plug>(coc-funcobj-i)
  omap if <Plug>(coc-funcobj-i)
  xmap af <Plug>(coc-funcobj-a)
  omap af <Plug>(coc-funcobj-a)
  ```

- Create file `main.cpp` with:
  ```cpp
  void f()
  {
  	// body (1)
  }
  auto g()
    -> void
  {
  	// body (2)
  }
  ```

- Start (neo)vim with command: `nvim -u mini.vim main.cpp`
- With the cursor on `(1)`, hit `vaf`. The selection spans from the beginning of the first line until the first slash; we would expect the selection to end one line later.
- With the cursor on `(1)`, hit `vif`. The selection spans from the brace in the second line until the first slash on line 3.
- With the cursor on `(2)`, hit `vif`. This time the selection includes the return type (`-> void`) and the entire line before the closing brace.

## Screenshots (optional)

coc-test coc-test2

marcdeop:
Fixes #1847

### DIFF
--- a/src/handler/symbols/index.ts
+++ b/src/handler/symbols/index.ts
@@ -152,8 +152,8 @@ export default class Symbols {
     if (inner && selectRange) {
       let { start, end } = selectRange
       let line = doc.getline(start.line + 1)
-      let endLine = doc.getline(end.line - 1)
-      selectRange = Range.create(start.line + 1, line.match(/^\s*/)[0].length, end.line - 1, endLine.length)
+      let endLine = doc.getline(end.line)
+      selectRange = Range.create(start.line + 1, line.match(/^\s*/)[0].length, end.line, endLine.length)
     }
     if (selectRange) {
       await window.selectRange(selectRange)


fannheyward:
This didn't fix #1847 for all languages, only works on Python with Pyright. I've tried several times with #1847, but still not find a better way to fix it.

coc uses LSP's `DocumentSymbol` to do function/class select, for example

```py
def foo():
    print(1)


class FooCog:
    def __init__(self) -> None:
        print(2)
```

Pyright returns three symbols (formatted), `Function foo`,  `Class FooCog`, `Method __init__`, with location range:

```json
[
    {
        "col": 5,
        "lnum": 1,
        "text": "foo",
        "level": 0,
        "kind": "Function",
        "range": {
            "start": { "line": 0, "character": 0 },
            "end": { "line": 1, "character": 12 }
        },
        "selectionRange": {
            "start": { "line": 0, "character": 4 },
            "end": { "line": 0, "character": 7 }
        }
    },
    {
        "col": 7,
        "lnum": 5,
        "text": "FooCog",
        "level": 0,
        "kind": "Class",
        "range": {
            "start": { "line": 4, "character": 0 },
            "end": { "line": 6, "character": 16 }
        },
        "selectionRange": {
            "start": { "line": 4, "character": 6 },
            "end": { "line": 4, "character": 12 }
        }
    },
    {
        "col": 9,
        "lnum": 6,
        "text": "__init__",
        "level": 1,
        "kind": "Method",
        "range": {
            "start": { "line": 5, "character": 4 },
            "end": { "line": 6, "character": 16 }
        },
        "selectionRange": {
            "start": { "line": 5, "character": 8 },
            "end": { "line": 5, "character": 16 }
        }
    }
]
```

when you do `coc-funcobj-i` on `print(1)`, coc will use current cursor position to find `Function foo`, with range `[0, 0, 1, 12]`, then increase `start.line +1` and decrease `end.line - 1` as inner range.

This works for languages that have end sign line, for example `}` for C/C++/Go/Rust/TS, `end` for Lua/Ruby, but for Python, it doesn't has special end sign, decrease the end.line can't include last line https://github.com/fannheyward/coc-pyright/issues/765. Your fix  will work for Python but not C/Go/TS etc.

Two conditions are needed for inner selection to work properly in `coc-funcobj-i`:

1. the code is formatted correctly
2. check function end sign, like `}` or `end`

The first condition:

```c
void f()
{
	// body (1)
}
```

inner selection on `body (1)` is incorrect because it contains `{`.

```c
void f() {
  // body (1)
}
```

this works as expected.

The second condition, for languages that have `}`, `end`, decrease `end.line` works as expected, but for Python, coc shouldn't decrease `end.line`.

I used to try check the language to decrease `end.line` or not, but I don't think this is a better solution.

marcdeop:
Apologies, I only read [this comment](https://github.com/neoclide/coc.nvim/issues/1847#issuecomment-623103768) and thought I would give it a shot.

Thanks for the explanation on the issue here.

If I understood your description correctly... this inner selection should also fail:
```
void f() {
  // body (1)}
```
right?   **note**:(notice the `}` after the body.

Possible solutions:
* coc is aware of the language and its particular way of defining function limits
* the language server includes information of those limits in its messages
* ... ?

For the sake of pragmatism and seeing that right now the most affected language is python... perhaps we should indeed deal with python differently and don't do the `-1` on it.

Opinions?

marcdeop:
I am not at Typescript developer and this is not a super easy change.

Closing PR before it lingers here forever.

PR #4526, opened 2023-02-20
fix bug: patchLine, byte aligned

solomonwzs:
modify function `patchLine`, let it filled character align by byte, avoid it causing bug at fuzzy match highlights like this:

image

### DIFF
--- a/src/util/diff.ts
+++ b/src/util/diff.ts
@@ -56,7 +56,7 @@ export function diffLines(oldLines: ReadonlyArray<string>, newLines: ReadonlyArr
 export function patchLine(from: string, to: string, fill = ' '): string {
   if (from == to) return to
   let idx = to.indexOf(from)
-  if (idx !== -1) return fill.repeat(idx) + from
+  if (idx !== -1) return fill.repeat(byteLength(to.substring(0, idx))) + from
   let result = fastDiff(from, to)
   let str = ''
   for (let item of result) {


fannheyward (approved):

fannheyward:
@solomonwzs thank you for your PR, can you post an screenshot without this PR? The code is working but I don't get the issue point.

solomonwzs:
> modify function `patchLine`, let it filled character align by byte, avoid it causing bug at fuzzy match highlights like this:
> 
> image

@fannheyward this screenshot was the screenshot without this PR. I input "ins" to do fuzzy match, but it didn't highlight the right positions.

Suppose I has a `ListItem` like this:

``` ts
{
    label: "hello world",
    filterText: "world",
}
```

It set `filterText` option, when we do fuzzy match, it would call `patchLine` to patch the lable, replace the filter text with whitespace avoid match it.

In this case:

```
original text: "hello world" 
patch text:    "      world"
```

It do fuzzy match at patch text and get the match position, then highlight original text with the position data.

If the original text has utf8 character which contain more then one byte, original code would only replace it with one whitespace:

```
original text: "中 world"
patch text:    "  world"
```

When I input `l`, it do fuzzy match at patch text and get position `5`, then highlight 5th character at original text, and length of `中` was 3, so it would highlight `o` but not `l`.

PR #4567, opened 2023-03-14
feat(core): don't resolve root for readonly file

fannheyward:
For example you're editing a rust file, `jumpDefinition` to std lib like `/opt/homebrew/Cellar/rust/1.68.0/lib/rustlib/src/rust/library/std/src/alloc.rs`, coc opens this document, will resolve root for it, then `didChangeWorkspaceFolders` notification to language server, this made the language server try to index and analyze the std libs. The std libs should be added to workspace in `didChangeWorkspaceFolders`.

This patch may not be a complete solution, for example rust std file is readonly, but go's not

`&readonly` of  `/opt/homebrew/Cellar/rust/1.68.0/lib/rustlib/src/rust/library/std/src/alloc.rs` is 1

`&readonly` of `/opt/homebrew/Cellar/go/1.20.2/libexec/src/context/context.go` is 0

Maybe a better way to detect std files and prevent coc to resolve root.

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -489,6 +489,7 @@ function! coc#util#get_bufoptions(bufnr, max) abort
         \ 'variables': coc#util#variables(a:bufnr),
         \ 'filetype': getbufvar(a:bufnr, '&filetype'),
         \ 'lisp': getbufvar(a:bufnr, '&lisp'),
+        \ 'readonly': getbufvar(a:bufnr, '&readonly'),
         \ 'iskeyword': getbufvar(a:bufnr, '&iskeyword'),
         \ 'changedtick': getbufvar(a:bufnr, 'changedtick'),
         \ 'fullpath': empty(bufname) ? '' : fnamemodify(bufname, ':p'),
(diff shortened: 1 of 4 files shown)


oxalica:
External crates like `use libc;` in Rust jumps to `$CARGO_HOME/registry/src/github.com-1ecc6299db9ec823/libc-0.2.140/src/lib.rs`, which is not `&readonly` but expected to be readonly by cargo and typical users.

fannheyward:
> which is not &readonly

Yes, this is the key issue, not all std files are made as readonly.

jackielii:
Same things happens to gopls. I had this problem for a long time... My solution is just disable the workspace folder resolution. Always set to pwd.

But somehow there are always some plugin trying to change my cwd and triggering the re-indexing. I just find them and disable them or remove them :D

fannheyward:
@jackielii set `workspace.ignoredFolders`

```json
  "workspace.ignoredFolders": [
    "$HOME",
    "$HOME/.cargo/**",
    "$HOME/.rustup/**",
    "$HOMEBREW_PREFIX",
    "$HOMEBREW_PREFIX/**"
  ],
```

luisdavim:
For Go there's `$GOPATH` but it's a `PATH` style var, so it can be a list of dir paths separated by `:` and in those cases we can use it directly to exclude in the configuration.

PR #4591, opened 2023-03-30
fix(outputChannel): avoid empty lines

fishioon:


fannheyward:
This breaks `appendLine`.

PR #4598, opened 2023-04-08
Disabled symbol loading message

blockchainian:
this pr removes the symbol loading message from the outline window.


fannheyward:
Thank you for your PR, but the loading message will be cleared after outline tree view  displayed, why do you need to disable it? Any screenshot?

blockchainian:
@fannheyward thanks for taking a look. i can't reproduce this any more.

when i was setting up CoC, this message did show up in the outline, so i removed it. i also ran into some timing bugs in my VIM config and have since added a few sleep calls to work around them. things now work like a blast!

i am closing this pr now.

PR #4601, opened 2023-04-10
feat: add coc_get_tag_timeout

gou4shi1:
To avoid `<C-]` stuck for a very long time.


fannheyward:
Thank you for your PR, this works!

I'd prefer to add the timeout configuration like `coc.preferences.tagDefinitionTimeout`, most coc.nvim configurations prefer to schema.json, especially used in TypeScript codes.

PR #4628, opened 2023-05-04
CocStart function added to readme before CocInstall

varolbora5:
I felt the need to specify that we need to do CocStart before being able to use any functions so I added it to the readme


fannheyward:
No, you don't need to `CocStart` to run `CocInstall`, coc.nvim will start up and run automatically.

PR #4629, opened 2023-05-05
Remove outdated keybinding from lua conf

samhed:
The binding for "coc-snippets-expand-jump" don't need to be set manually anymore. These bindings are set up automatically when snippets are available now-a-days.


fannheyward:
Thank you for your PR, but `coc-snippets-expand-jump` is used to `expand` snippet and `jump` to next placeholder, it's provided by coc-snippets.

samhed:
Ok, sorry I misunderstood things.

It's a bit strange that the vimscript example config doesn't mention `coc-snippets-expand-jump` anymore. Otherwise, the two example-configs seem very similar.

fannheyward:
The `coc-snippets-expand-jump` vimsctipt is in coc-snippets README.

PR #4648, opened 2023-05-23
feat(inlayHint): inline display for nvim

fannheyward:
needs https://github.com/neovim/neovim/pull/20130

Next:

- [x] remove the separator, this is a breaking change, but we don't needs separator for inline inlayHint, same as vim
    - inlayHint.subSeparator
    - inlayHint.typeSeparator
    - inlayHint.parameterSeparator
- [x] clean docs of separator
- [x] tests // The CI runs with nvim 0.9.0, which didn't support inline virt, works on my local

Do we need to support both `inline` for nvim 0.10.0+ and `eol` for nvim 0.9.0-? I do like only `inline`, for nvim 0.9.0 and before, no inlayHint

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -1304,29 +1304,11 @@
         "type": "string"
       }
     },
-    "inlayHint.parameterSeparator": {
-      "type": "string",
-      "scope": "language-overridable",
-      "default": "",
-      "description": "Separator for parameter inlay hint, neovim only."
-    },
     "inlayHint.refreshOnInsertMode": {
       "type": "boolean",
       "default": false,
       "scope": "language-overridable",
       "description": "Refresh inlayHints on insert mode."
-    },
-    "inlayHint.subSeparator": {
-      "type": "string",
-      "scope": "language-overridable",
-      "default": " ",
-      "description": "Separator for chained inlay hints, neovim only."
-    },
-    "inlayHint.typeSeparator": {
-      "type": "string",
-      "scope": "language-overridable",
-      "default": "",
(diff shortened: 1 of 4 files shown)


judecumt:
Finally, Today is a great day!

oblitum:
Hi @fannheyward

> Do we need to support both `inline` for nvim 0.10.0+ and `eol` for nvim 0.9.0-? I do like only `inline`, for nvim 0.9.0 and before, no inlayHint

I agree with this, the `eol` behavior is really not worth the end result, nor the complexity to keep maintaining it.

All the checkboxes are ticked, but the current CI configuration picks last nvim stable, which is 0.9.0. Do you assume it's OK to merge now despite any failing tests on CI (if there's any)? Or is it better to wait nvim 0.10 first?

oblitum:
Just pulled nvim-nightly-bin from AUR and tested the branch on my machine, working nicely. Had to rebase over master due to new node 20 support after upgraded dev-deps.

oblitum:
Hi @fannheyward. After last rebase do you expect review for merge? I gave a review last time but I think my questions still stand and are still unanswered.

oblitum:
@fannheyward ok. [This](https://github.com/neoclide/coc.nvim/pull/4648/files#diff-d199ecad7c8c0e2aa4b5c28dd8cf81fc349edcd523c29b59927b53b17a25f97aL190) is pending, not sure whether you wish to leave that for later too.

chemzqm:
Works as expected for me.

PR #4676, opened 2023-06-16
Remove <c-j> in example lua config

Summon528:
Not sure why this config is in there. It's not in the vim script version and it literally just types `<Plug>(coc-snippets-expand-jump)`  on my machine. Remove it restore the snippet jump function since `c-j` is already the default.


fannheyward:
https://github.com/neoclide/coc.nvim/pull/4629

PR #4722, opened 2023-08-05
fix: ls provider's score is equal to the default one

tcx4c70:
For now, ls provider's score will be equal to the default one if ls provider's documentSelector has pattern. For example, a ls registers capability `textDocument/documentHighlight` with documentSelector `{'language': 'cs', 'scheme': 'file', 'pattern': '**/*.cs'}`, it will get score 5, which is equal to the default one with documentSelector '*', even if the language id has matched. Since the capability is registered later than the default one, DocumentHighlightManager will only use the default one to provide highlight, which is not expected.

If all of language, scheme, and pattern match, we should return the highest score of them, instead of the last score.


fannheyward:
in testing, sorry for the delay review.

fannheyward:
Please test for this case:

`expect(workspace.match([{ language: 'xml', scheme: 'file', pattern: '**/*.xml' }], doc.textDocument)).toBe(10)`

to `src/__tests__/modules/workspace.test.ts` L295

fannheyward (approved):

PR #4741, opened 2023-09-01
feat: rm yarn, use npm

fannheyward:
Remove `yarn` totally, use `npm` instead.

After we upgrade `glob` to v10 https://github.com/neoclide/coc.nvim/commit/34f5d9671b7ddfe1d7a3d4fdbe89684432f1b2af, this uses `jackspeak` , and `cliui`, which uses an renamed `string-width` that broken testing work. https://github.com/isaacs/jackspeak/issues/5

To fix this issue, use an earlier `glob`, or upgrade to yarn 3. This won't be fixed with yarn 1.

I've tested several ways yesterday:

1. remove yarn 1, use `npm`.
    - pros: yarn 1 is dead, long live npm
    - cons: breaking change for `master` users
2. use `glob@9`, also supports signal that to fix https://github.com/neoclide/coc.nvim/issues/4538
    - pros: no `jackspeak`, no breaks for anything
    - cons: lock to `glob@9`
3. we've upgraded node to v16.18, we can use `corepack` to use yarn 3: `corepack enable; corepack prepare yarn@3.6.3 --activate`
    - pros: still use yarn, no more yarn 1 issues
    - cons: needs steps to enable yarn with corepack
4. I also tested with `pnpm` because more frontend projects like vue uses this, we can also use `corepack` to enable this
    - pros: no more yarn 1 issues, more modernize? 
    - cons: needs steps to enable it, change our CI, development

Because yarn 1 is dead, I prefer to use `npm`, remove all yarn scripts.

https://github.com/neoclide/coc.nvim/commit/15c894666d5d73236536fb1ae017e8d87726c246#commitcomment-126134192


fannheyward:
cc @chemzqm @yaegassy

fannheyward:
User can still use `yarn` or `cnpm` with `npm.binPath` to install extensions.

Issue #4755, opened 2023-09-12
Hover doesn't work in nimlangserver


futureisfoss:
## Result from CocInfo

```
## versions

vim version: NVIM v0.9.1
node version: v20.5.1
coc.nvim version: 0.0.82-3dc6153a 2023-07-21 00:22:17 -0500
coc.nvim directory: /home/rokosun/.config/nvim/plugged/coc.nvim
term: st-256color
platform: linux

## Log of coc.nvim

2023-09-12T16:13:28.516 INFO (pid:30989) [plugin] - coc.nvim initialized with node: v20.5.1 after 664
2023-09-12T16:13:28.533 INFO (pid:30989) [services] - LanguageClient nim state change: stopped => starting
2023-09-12T16:13:28.560 INFO (pid:30989) [language-client-index] - Language server "languageserver.nim" started with 31012
2023-09-12T16:13:28.616 INFO (pid:30989) [services] - LanguageClient nim state change: starting => running
2023-09-12T16:13:28.648 INFO (pid:30989) [services] - service languageserver.nim started
2023-09-12T16:13:38.065 INFO (pid:30989) [attach] - Request action: hasProvider [ 'hover' ]
2023-09-12T16:13:38.073 INFO (pid:30989) [attach] - receive notification: doHover []
```

## Describe the bug

The hover works fine for me in [coc-sh](https://github.com/josa42/coc-sh) and [coc-clangd](https://github.com/clangd/coc-clangd) but for some reason it doesn't work with [nimlangserver](https://github.com/nim-lang/langserver), when I try to Hover on Nim files it just says `[coc.nvim] hover not found` at the bottom. One thing to note here is that the completions and everything else works fine for me in Nim, just the Hover functionality is missing. I've also made an issue on the [nimlangserver's git repo about this](https://github.com/nim-lang/langserver/issues/47) and they've suggested me to discuss it here, I suspect there might be some compatibility issues going on.

## Reproduce the bug

- Install and setup `nimlangserver` like [explained here](https://github.com/nim-lang/langserver#vimneovim)

- Create file `mini.vim` with：

  ```vim
  set nocompatible
  set runtimepath^=/path/to/coc.nvim
  filetype plugin indent on
  syntax on
  set hidden
  
  " Use K to show documentation in preview window
  nnoremap <silent> K :call ShowDocumentation()<CR>
  
  function! ShowDocumentation()
    if CocAction('hasProvider', 'hover')
      call CocActionAsync('doHover')
[...]

PR #4757, opened 2023-09-13
fix(hover): rm hover validation

fannheyward:
https://github.com/neoclide/coc.nvim/issues/4755, some LS responses with invalid `hover.range`, remove this validation doesn't break hover actions:

1. hover.range is only used to highlight the hover object, this will be checked later
2. previewHover will check hover.contents


chemzqm:
Should be fixed by eded814d

PR #4758, opened 2023-09-13
Allow configuration of the debounce time on current function setting

3ximus:
And also changed the docs in accordance to the new changes to cursorMove event


3ximus:
Fixed issues with stable tests. If you accept it should be good to merge

PR #4810, opened 2023-11-13
fix: CocTagFunc jumps to wrong position if using tabs for indentation

tcx4c70:
From LSP specification, Position.character indicates the character offset on a line in a document (zero-based) [1]. However, `|` in vim will jump to screen column [2]. If the file (e.g., linux kernel source code) uses tabs for indentation, then CocTagFunc will jump to a wrong position. We can fix it by changing to use `cursor()`, whose col argument means the byte-count column.

1. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
2. https://vimdoc.sourceforge.net/htmldoc/motion.html#bar

### DIFF
--- a/src/__tests__/handler/locations.test.ts
+++ b/src/__tests__/handler/locations.test.ts
@@ -278,10 +278,10 @@ describe('locations', () => {
       expect(res).toEqual([
         {
           name: 'foo',
-          cmd: 'silent keepjumps 3 | normal 1|',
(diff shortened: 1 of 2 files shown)


fannheyward (approved):

fannheyward:
@all-contributors please add @tcx4c70 for code.

Issue #4879, opened 2024-02-02
vim very slow with semantic tokens enabled (with a possible fix)


cridemichel:
## Result from CocInfo

```
## versions

vim version: VIM - Vi IMproved 9.1 9010050
node version: v21.6.1
coc.nvim version: 0.0.82-d1568d56 2023-09-29 19:43:34 +0800
coc.nvim directory: /Users/demichel/.vim/plugged/coc.nvim
term: iTerm.app
platform: darwin

## Log of coc.nvim

2024-02-02T16:25:43.166 INFO (pid:63048) [plugin] - coc.nvim initialized with node: v21.6.1 after 245
2024-02-02T16:25:43.868 INFO (pid:63048) [attach] - receive notification: showInfo []
```
## VIM VERSION

```
:version
:version
VIM - Vi IMproved 9.1 (2024 Jan 02, compiled Jan 23 2024 22:19:02)
macOS version - x86_64
Included patches: 1-50
Compiled by Homebrew
Huge version without GUI.  Features included (+) or not (-):
+acl               +comments          +find_in_path      +lispindent        +multi_byte        +rightleft         +terminal          +wildignore
+arabic            +conceal           +float             +listcmds          +multi_lang        +ruby              +terminfo          +wildmenu
+autocmd           +cryptv            +folding           +localmap          -mzscheme          +scrollbind        +termresponse      +windows
+autochdir         +cscope            -footer            +lua               +netbeans_intg     +signs             +textobjects       +writebackup
-autoservername    +cursorbind        +fork()            +menu              +num64             +smartindent       +textprop          -X11
-balloon_eval      +cursorshape       +gettext           +mksession         +packages          +sodium            +timers            -xattr
+balloon_eval_term +dialog_con        -hangul_input      +modify_fname      +path_extra        +sound             +title             -xfontset
-browse            +diff              +iconv             +mouse             +perl              +spell             -toolbar           -xim
++builtin_terms    +digraphs          +insert_expand     -mouseshape        +persistent_undo   +startuptime       +user_commands     -xpm
+byte_offset       -dnd               +ipv6              +mouse_dec         +popupwin          +statusline        +vartabs           -xsmp
+channel           -ebcdic            +job               -mouse_gpm         +postscript        -sun_workshop      +vertsplit         -xterm_clipboard
+cindent           +emacs_tags        +jumplist          -mouse_jsbterm     +printer           +syntax            +vim9script        -xterm_save
[...]

Issue #4872, opened 2024-01-27
error messages if semanticTokens are enabled


cridemichel:
## versions

vim version: VIM - Vi IMproved 9.1 9010050
node version: v21.6.1
coc.nvim version: 0.0.82-d1568d56 2023-09-29 19:43:34 +0800
coc.nvim directory: /Users/demichel/.vim/plugged/coc.nvim
term: iTerm.app
platform: darwin

## Log of coc.nvim

2024-01-27T18:46:11.210 INFO (pid:38752) [plugin] - coc.nvim initialized with node: v21.6.1 after 518
2024-01-27T18:46:11.236 INFO (pid:38752) [services] - LanguageClient clangd state change: stopped => starting
2024-01-27T18:46:11.241 INFO (pid:38752) [language-client-index] - Language server "clangd" started with 38754
2024-01-27T18:46:11.441 INFO (pid:38752) [services] - LanguageClient clangd state change: starting => running
2024-01-27T18:46:11.451 INFO (pid:38752) [services] - service clangd started
2024-01-27T18:46:14.656 INFO (pid:38752) [attach] - receive notification: showInfo []

## coc-settings.json
```json
{
  "diagnostic.refreshOnInsertMode": true,
  "semanticTokens.enable": true,
  "diagnostic.checkCurrentLine": true,
  "suggest.enablePreview": true,
  "suggest.triggerAfterInsertEnter": true,
  "diagnostic.messageTarget": "float",
  "diagnostic.enable": false,
  "inlayHint.enable" : true,
  "suggest.noselect": true,
  "suggest.minTriggerInputLength": 2,
  "inlayHint.enableParameter": false,
  "coc.source.vimtex.priority": 99,
  "python.pythonPath": "/usr/local/bin/python3",
  "markdown-preview-enhanced.enableScriptExecution": true,
  "coc.preferences.diagnostic.displayByAle": true,
  "list.source.files.args": ["--hidden", "--files"],
  "yank.highlight.enable": false,
  "pyright.inlayHints.parameterTypes": false,
  "pyright.inlayHints.variableTypes": false,
  "pyright.disableDocumentation": false,
  "pyright.enable": true,
  "semanticTokens.enable": true,
  "colors.enable": true,
  "links.enable": false,
  "links.highlight": false,
  "links.tooltip": false,
  "suggest.floatConfig": {"border": false, "rounded": false}
}
```

## Describe the bug

vim produces several error messages by jumping back and forth between two files
[...]

PR #4885, opened 2024-02-03
fix issues #4879 and #4872

cridemichel:
**this PR fixes issues #4879 and #4872**
[#4879] list slicing is used in the function "s:add_highlights_timer", instead of a very slow for loop
[#4872] ignore error E964


fannheyward (approved):

fannheyward:
@all-contributors please add @cridemichel for code.

PR #4929, opened 2024-03-07
fix(list): set foldcolumn to 0 since it isn't applicable

mehalter:
If the user uses `foldcolumn` then the list buffers have an extra column that is not useful especially for a buffer that is bound to only be in insert mode. This locally sets `foldcolumn=0`


mehalter:
Not sure what's happening with the tests here, seems unrelated to the chance made in the PR 🤔  Let me know if there's anything I should look into! Thanks 😄

fannheyward (approved):

fannheyward:
@all-contributors please add @mehalter for code.

PR #4949, opened 2024-03-20
fix(inlayHint): hl_mode = 'replace' for neovim

kevinhwang91:
Problem: https://github.com/neovim/neovim/issues/24152 The inlays are affected by the other highlight groups.
Solution: use `replace` value for `hl_mode` instead of  `combine`.


fannheyward:
From `:h nvim_buf_set_extmark`, the `hl_mode`'s default value is `replace`, did I misunderstand it?

PR #4950, opened 2024-03-20
fix(inlayHint): more compact padding highlight group

kevinhwang91:
If the chunk is an array with a single string, the highlight group only uses the foreground of `Normal`, which makes `cursorline` look better.

Reference: https://github.com/neovim/neovim/blob/0f85aeb478a68c18a8195e69724e1cb618fec058/runtime/lua/vim/lsp/inlay_hint.lua#L333-L337


qiqiboy:
inlayhint in vim9 not work from this commit

fannheyward:
> inlayhint in vim9 not work from this commit

Reproduced

fannheyward:
@qiqiboy made a quick fix, set hl_group for hint in Vim

PR #4954, opened 2024-03-23
fix(defaultAction) fallback to default action on emtpy expression

asmodeus812:
This should allow the expression function to return nil, to by default execute the default action of the list, as there are no easy ways to extract the name of the default action for the current list. doAction by default if no 'name' is provided would fallback to the defaultAction.

### DIFF
--- a/src/list/mappings.ts
+++ b/src/list/mappings.ts
@@ -141,7 +141,7 @@ export default class Mappings {
     })
     this.addAction('expr', async expr => {
       let name = await manager.call(expr)
-      if (name) await manager.doAction(name)
+      await manager.doAction(name)
     })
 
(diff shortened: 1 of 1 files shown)


asmodeus812:
The reason is to be able to do something like that

```lua
function _G.list_default_action(ctx)
        if ctx and #ctx.targets > 1 then
            -- send to qflist
            return "quickfix"
        else
            -- default
            return nil
        end
    end
```

```viml
        function! Listaction(ctx)
            " do execute quickfix or default action
            return v:lua.list_default_action(a:ctx)
        endfunction
```

```json
   "<CR>": "expr:Listaction",
````

fannheyward (approved):

PR #4961, opened 2024-03-28
chore(CI): use Python 3.12

cclauss:
Python 3.12 can cause issues because `distutils` was removed from the Standard Library.


fannheyward:
@all-contributors please add @cclauss for code.

PR #4985, opened 2024-04-19
fix(spawn): add `shell:true` on windows

fannheyward:
Closes https://github.com/neoclide/coc-prettier/issues/178

Latest version of Node.js needs to set `shell: true` to spawn option

- https://github.com/node-red/node-red/pull/4652
- https://github.com/microsoft/rushstack/pull/4637


tankorsmash:
I had a similar issue with running the Elm Language Server via `.cmd` and Node 20. Updating coc to a version with this fix (latest version) solved the issue. Thank you so much!

PR #4997, opened 2024-04-28
fix(list): update lastSession when resume list by name

### LINKED ISSUE #4971: List not work after some operation
## Result from CocInfo

The `CocInfo` seems broken, I don't know why. 
coc.nvim is the latest master. 
vim 9.0.1563/ nvim v0.9.0
> [coc.nvim]: Error on notification "showInfo": stripAnsi3 is not a function

## Describe the bug

Some keymaps not work in list window. `Enter` do not open the file (for files list), `TAB` no not show the action.

## Reproduce the bug

case 1:

- Type `:CocList files` to open files list.
- Type `Ctrl-o` to switch to NORMAL model.
- Type `:` and then `<ESC>`.
- Press `<Enter>` and can't open the file under cursor.

case 2:

- `:CocList files` to open files list and then open one file.
- `:CocList grep int` to open grep list and then open one file location.
- `:CocListResume files`, can't open files.

## Screenshots (optional)

NULL

aispeaking:
List action works on lastSession. When resuming a named list which is not the lastSession, it should update the lastSession.

This fix #4971 case2.

### DIFF
--- a/src/list/manager.ts
+++ b/src/list/manager.ts
@@ -154,6 +154,7 @@ export class ListManager implements Disposable {
         void window.showWarningMessage(`Can't find exists ${name} list`)
         return
       }
+      this.lastSession = session
       await session.resume()
     }
   }


fannheyward (approved):

fannheyward:
@all-contributors please add @aispeaking for code.

PR #5002, opened 2024-05-02
feat(extensions): allow github extensions to be installed through development mode

asmodeus812:
Problem description https://github.com/neoclide/coc.nvim/discussions/5001


fannheyward:
Similar discussion https://github.com/neoclide/coc.nvim/discussions/4960

I'd like to recommend to use plugin manager to install extensions that not published to npm yet. Something like `Plug 'neoclide/coc-tsserver', {'do': 'yarn install --frozen-lockfile'}`  https://github.com/neoclide/coc.nvim/wiki/Using-coc-extensions#use-vims-plugin-manager-for-coc-extension

This fix treats the URL `github.com` as whitelist to allow running in development mode, what if an extension hosted with `gitlab.com` or `sr.ht` or custom git repo? My thought: limit `CocInstall` to run in production only, and use plugin manger to run in development mode, with more flexibility.

asmodeus812:
Hi Thanks for taking a look, 

The main reason i would like to have coc manage this instead of a plugin manager, is first most package managers do not provide an api to do this programatically, they are declarative, so you have do it in your plugin config, where as i want the clone and build of the extension to happen programatically when i choose that. I can easily do that with `CocInstall` for regular npm hosted packages. I do not like the idea of having two completely separate paths working differently but essentially doing the same thing. 

Secondly there is no way to ever build an extension from a clone of a `github` repo with the current implementation. Most of coc extensions have a lot of devDependencies, and the current implementation fails to build everthing that it downloads from github, due to the `devDeps` not being downloaded in `production` mode. Meaning that feature (to pull an extension from github) is effectively `dead` as it is. It is not working anyway. 

As for the source being github, can be easily extended to work with other hosts, but most package managers only support github (maybe gitlab too, at the very best) to begin with any way. 

If it is a custom git repo, it should be probably managed by the user, i think coc providing a way to clone from github is at the very least an MVP, anything more should be managed by the user.

I just wanted to make the existing feature `work` correctly.

fannheyward:
cc @Freed-Wu

Issue #5028, opened 2024-05-31
Asynchronous Completion for `isIncomplete` Items in coc.nvim


hexh250786313:
In the current implementation of coc.nvim, when there are `isIncomplete` items in the completion panel, the entire panel is blocked waiting for the completion results of these `isIncomplete` items. During this time, even if the user has other inputs, the completion panel will not update because it is completely blocked. 

This behavior is not very user-friendly, as the completion results for `isIncomplete` items may be slow, leading to a poor experience with the entire completion panel. I suggest that for these `isIncomplete` items, completion should be performed asynchronously in the background. The completion panel can then be updated once their completion results are returned, rather than blocking the entire panel. 

Meanwhile, other non-`isIncomplete` items should be able to display their results immediately, instead of being blocked along with the `isIncomplete` items.

https://github.com/neoclide/coc.nvim/assets/26080416/246ac5ae-c0bc-435b-ba82-4c039856ee16

As can be seen from the screenshot above, when I input `s` and remove it, the entire completion panel updates very slowly, it is completely blocked by the `[Cop]` completion source. In this example, the `[A]` completion source should not wait for the results of `[Cop]`, it should update immediately when `s` changes, and then update the entire completion panel again after the results of `[Cop]` are returned.

PR #5029, opened 2024-05-31
Fix Completion Blocked by `incomplete` Item

hexh250786313:
Should fix https://github.com/neoclide/coc.nvim/issues/5028.

**This pull request is still under testing.**

### DIFF
--- a/src/completion/complete.ts
+++ b/src/completion/complete.ts
@@ -241,11 +241,10 @@ export default class Complete {
     return added
   }
 
-  public async completeInComplete(resumeInput: string): Promise<DurationCompleteItem[] | undefined> {
+  public async completeInComplete(resumeInput: string) {
     let { document } = this
     this.cancelInComplete()
     let tokenSource = this.createTokenSource(true)
-    let token = tokenSource.token
     await document.patchChange(true)
     let { input, colnr, linenr, followWord, position } = this.option
     Object.assign(this.option, {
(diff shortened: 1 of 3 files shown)


src/completion/complete.ts
```diff
@@ -257,11 +256,8 @@ export default class Complete {
       triggerCharacter: undefined,
       triggerForInComplete: true
     })
-    this.cid++
     const sources = this.getIncompleteSources()
     await this.completeSources(sources, tokenSource, this.cid)
-    if (token.isCancellationRequested) return undefined
-    return this.filterItems(resumeInput)
```
fannheyward, line 264: I haven't tested this yet, but will this change prevent `incomplete` being cancelled?

hexh250786313: @fannheyward 
Good point! I haven't tested this point yet either, but there is the following code in `this.completeSources(...)`. I believe it will decide whether to refresh the completion panel based on the value of `isCancellationRequested`:

https://github.com/neoclide/coc.nvim/blob/7adbccbe8fa08a1c5031f62f8d2c6296f49e94b8/src/completion/complete.ts#L146-L149

I will test it shortly. If there are any issues, fix will coming.

hexh250786313:
https://github.com/neoclide/coc.nvim/blob/7adbccbe8fa08a1c5031f62f8d2c6296f49e94b8/src/completion/pum.ts#L175-L177

It defaults to setting the first item as the selected state. For `incompletion` items with high priority values, they become the top item every time after refresh, causing them to be selected by default even if `preselect` is not configured. I must say, this behavior is a bit annoying. Need to fix.

chemzqm:
Should be fixed by 2c067b2e

Issue #4740, opened 2023-08-31
Inlay hint configurable option


lmburns:
**Is your feature request related to a problem? Please describe.**
I do not like the inline inlay hints.

**Describe the solution you'd like**
Make it a configuration option on whether or not separators are used, or where the inlay hint is located. Similar to how the code-lens option is given to choose.

**Describe alternatives you've considered**
Disabling them, but I'd prefer not to.

**Additional context**

chemzqm:
Separator of inlay hint have removed, not needed any more, see https://github.com/neoclide/coc.nvim/blob/master/history.md#2023-08-31

lmburns:
Yeah I saw that it was removed. That's why I created this feature request. But I couldn't find out how to place the inlay-hints at the end of the line like they were before, instead of them mandatorily being inline. Is there a way to do this? I searched through the documentation

chemzqm:
> But I couldn't find out how to place the inlay-hints at the end of the line like they were before.

No such support as we don't think it's a good idea compare to inline ext markers.

SeniorMars:
I'm sure you have your reasons, but I would still like to request this feature. I'm not used to inline markers at all, and would prefer the old style.

Awpteamoose:
It's an opensource project and as authors you can do whatever you want, but I don't understand why are you so eager to break established habits of your users. It was the same with the floating status notifications around a year ago.

SeniorMars:
No, thank you for your work on this awesome project!

aemonge:
Just for keeping track on the other issue:

> Can we put the inline hints only visible for current line?

Ref. https://github.com/neoclide/coc.nvim/issues/4074?notification_referrer_id=NT_kwDOABQtbLI0MjI4NTE2NTU5OjEzMjIzNDg#issuecomment-2150493218

awused:
You can get the old hints by pinning 3dc6153.

Neovim 0.10.0 was only released a month ago, so I only just updated coc, but after seeing these new inlay hints I had to roll it back to 3dc6153 immediately.

fannheyward:
#5060 added `inlayHint.position` configuration, supports `inline` and `eol`.

fannheyward:
For the separator configuration, some language server returns with separator but some not, it's complex for the client to control separators, have no plan to add this.

Close as completed.

PR #5060, opened 2024-06-27
feat(inlayHint): add `inlayHint.position` configuration

de-vri-es:
This PR adds an option to move all inlay hints to *after* the original text.

Personally, I find hints inside the source code very distracting. And they make the code look much wider than it really is, which I personally do not like.

This option allows to restore the old behavior where all hints were displayed at the end of a line.

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -1303,6 +1303,13 @@
       "scope": "language-overridable",
       "description": "Enable inlay hint support"
     },
+    "inlayHint.position": {
+      "type": "string",
+      "default": "inline",
+      "scope": "language-overridable",
+      "description": "Controls where to show inlay hints: inline in the text, or at the end of the line",
+      "enum": ["inline", "eol"]
+    },
     "inlayHint.enableParameter": {
       "type": "boolean",
       "scope": "language-overridable",
--- a/src/handler/inlayHint/buffer.ts
+++ b/src/handler/inlayHint/buffer.ts
@@ -16,10 +16,16 @@ import workspace from '../../workspace'
 
 export interface InlayHintConfig {
   enable: boolean
+  position: InlayHintPosition,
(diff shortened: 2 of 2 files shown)


fannheyward:
Thank you for your PR! Please change the configuration to `inlayHint.position`, with `inline` and `eol` options.

https://github.com/neoclide/coc.nvim/issues/4740

de-vri-es:
Switched to use `config` instead of `_config` directly, should fix the test failure.

de-vri-es:
Hmm, can the test failure on nightly be related to this PR, or is it flaky? :o

fannheyward (approved):

de-vri-es:
Thanks! :partying_face:

SeniorMars:
I know this is a lot to ask but we used to be able to add an icon and color of the inlay hints in our cocconfig. Would it be possible to support those features as well?

de-vri-es:
You can configure the color by changing the `CocInlayHint`, `CocInlayHintType` and/or `CocInlayHintParameter` syn/highlight groups (from `:help coc-inlayHint`).

Icon I dunno.

SeniorMars:
I should probably make this an issue, but I'm not sure we want to support this.

In either case before (a long time ago), we used to have the following options.

```
 "codeLens.separator": " ",
  "codeLens.subseparator": " | ",
  "inlayHint.typeSeparator": "‣ ",
  "inlayHint.parameterSeparator": "‣ ",```

fannheyward:
@SeniorMars the separator configuration has been removed and has no plan to add back, because language server will add separator like `:` with inlay hint, we don't want to do omitting/replacing work in client side.

fannheyward:
@all-contributors please add @de-vri-es for code.

PR #5067, opened 2024-06-28
fix(coc#has_item_selected)

Kaiser-Yang:
I find that `coc#pum#visible` use `getwinvar(s:pum_winid, 'float', 0) == 1` to check the window. So using `s:pum_winid == -1` may cause problems, now use `coc#pum#visible()` to check if the `pum` is visible.


fannheyward:
@all-contributors please add @Kaiser-Yang for code.

PR #5102, opened 2024-08-03
Add `coc.preferences.formatterExtension` configuration

### LINKED ISSUE #5076: New configuration like VSCode's editor.defaultFormatter

It is often the case that different projects require different formatters for the same language. For example, in Python there are autopep8, black, isort, ruff, and many others. There are currently two solutions to this problem.

1. Support multiple formatters in an extension and switch formatters as an extension-specific setting. This is the approach taken by coc-pyright.
2. Provide a configuration to specify the priority of the formatter in the extension's own configurations and adjust the priority well with the other formatters. This is the approach taken by coc-prettier.

However, both of these have some issues. For 1, the extension must support all ever-existing formatters for that language, and this is not practical (and sometimes I use coc-diagnostic to run standalone formatters.) For 2, it is difficult to determine the single perfect priority value for an extension, especially for ones that support multiple file types (just as coc-prettier and coc-diagnostic.) For example, consider the situation where you want to use coc-prettier over coc-diagnostic in JavaScript, but want to use coc-diagnostic over coc-prettier in JSON. coc-prettier has its own configuration to disable formatter per language to support such a usecase, but I think that should be handled by coc.nvim's side.

To solve these problems, I want to have a new language-scope setting as VSCode's editor.defaultFormatter.

I initially tried to create a new extension to handle this (namely coc-formatter-dispatcher). If I could execute a specific formatter provided by other extensions, then registering the dispatcher with very high priority and dispatching format request to actual formatter solves the problem. But that was too difficult because there's no coc.nvim API to call another extension's formatter.

**Additional context**

statiolake:
This is a proof-of-concept.

- Registered by extensions itself (like `coc-pyright`): identified by extension name as expected
- Registered by extensions through a language client (like `@yaegassy/coc-ruff`): identified by extension name as expected
- Registered by coc.nvim for raw LSP `languageserver` configuration: identified as `coc.nvim`
  - TODO: So there is no way to select one from multiple LSP - what it should be in this case?

Priorities are still expected to work (when `coc.preferences.formatterExtension` is null).

Closes #5076

### DIFF
--- a/data/schema.json
+++ b/data/schema.json
@@ -630,6 +630,12 @@
         "type": "string"
       }
     },
+    "coc.preferences.formatterExtension": {
+      "type": ["null", "string"],
+      "default": null,
+      "scope": "language-overridable",
+      "description": "Extension used for formatting documents, defaulting to the formatter with highest priority."
+    },
     "coc.preferences.jumpCommand": {
       "anyOf": [
         {
--- a/data/schema.json
+++ b/data/schema.json
@@ -634,7 +634,7 @@
       "type": ["null", "string"],
       "default": null,
       "scope": "language-overridable",
-      "description": "Extension used for formatting documents, defaulting to the formatter with highest priority."
+      "description": "Extension used for formatting documents. When set to null, the formatter with highest priority is used."
     },
     "coc.preferences.jumpCommand": {
       "anyOf": [
--- a/doc/coc-config.txt
+++ b/doc/coc-config.txt
@@ -1624,6 +1624,12 @@ Preferences~
 
 	Scope: `language-overridable`, default: `false`
 
+"coc.preferences.formatterExtension"			*coc-preferences-formatterExtension*
+
+	Extension used for formatting documents. When set to null, the formatter with highest priority is used.
+
+	Scope: `language-overridable`, default: `null`
+
 "coc.preferences.jumpCommand"				*coc-preferences-jumpCommand*
 
 	Command used for location jump, like goto definition, goto references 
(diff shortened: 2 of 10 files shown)


fannheyward:
I've tested this and works as expected, thank you for your PR, the `parseExtensionName` function is a gem.

fannheyward:
Please add the update to `history.md` and `typings/index.d.ts` too.

statiolake:
Thank you for your checking!

For index.d.ts, is it correct understanding that it is listing public APIs of coc.nvim?
If so, it is a good option to keep the public API as-is to "hide" this extra implementation detail; an extension name is not what should be explicitly specified from an extension's side. What do you think? If you think we should include it in index.d.ts, I'll do that. And if to keep the current one, I'll add an explanation to the implementation side why this signature is different from the index.d.ts one.

Also, should I add some tests for this? The problem is that I'm not familiar with neither JavaScript's testing ecosystem nor coc.nvim's testing utilities.

fannheyward:
OK, let's _hide_ this from the API as the extension name won't be specified.

statiolake:
OK, then I won't update index.d.ts and ignore tests for now. But I think I still need to add this new configuration to history.md. I'll do that later today.

statiolake:
All done. Please check and merge this if it's OK.

fannheyward:
`doc/coc-config.txt` please

statiolake:
Sorry I didn't notice that. Now fixed.

fannheyward (approved):

PR #5114, opened 2024-08-16
Feature/expose more properties quick list

asmodeus812:
Added some more properties from the quick list to the typings to allow the user to have more contextual information about the ui element and it's state


fannheyward (approved):

fannheyward:
@all-contributors please add @asmodeus812 for code.

PR #5118, opened 2024-08-20
fix: ignore failed config file reads

### LINKED ISSUE #5117: `.vim` directory lookup across WSL can result in errors
## Result from CocInfo

```
## versions

vim version: VIM - Vi IMproved 9.0 9000000
node version: v21.7.1
coc.nvim version: 0.0.82-b7375d5f 2023-01-30 05:09:03 +0800
coc.nvim directory: H:\vim\plugged\coc.nvim
term: undefined
platform: win32

## Log of coc.nvim

2024-08-20T12:56:49.593 INFO (pid:11140) [plugin] - coc.nvim initialized with node: v21.7.1 after 418
2024-08-20T13:49:23.122 ERROR (pid:11140) [server] - unhandledRejection  Promise {
  <rejected> Error: ELOOP: too many symbolic links encountered, open '\\wsl.localhost\Ubuntu\home\j\.vim\coc-settings.json'
      at Object.readFileSync (node:fs:455:20)
      at hp.parseConfigurationModel (H:\vim\plugged\coc.nvim\build\index.js:157:4404)
      at hp.addFolderFile (H:\vim\plugged\coc.nvim\build\index.js:157:5286)
      at hp.locateFolderConfigution (H:\vim\plugged\coc.nvim\build\index.js:157:7909)
      at bp._createDocument (H:\vim\plugged\coc.nvim\build\index.js:177:9717)
      at H:\vim\plugged\coc.nvim\build\index.js:177:9186
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
    errno: -4067,
    code: 'ELOOP',
    syscall: 'open',
    path: '\\\\wsl.localhost\\Ubuntu\\home\\j\\.vim\\coc-settings.json'
  }
```

## Describe the bug

coc currently tries to resolve the closest `.vim` directory here:

https://github.com/neoclide/coc.nvim/blob/98fa66ae6a0bca9d1a9ab9bf5c6fe378aaa53f11/src/configuration/index.ts#L383-L397

presumably so it can later find a `coc-settings.json`.

if you're on a windows machine, using a windows version of vim to edit files within WSL (linux), this code will resolve to the linux `.vim` in some cases which isn't really what you want.

in this case, we hit some symlink nonsense because windows can't understand the ubuntu symlinks (it is indeed a symlink but you can read it fine in linux/wsl, it isn't a loop). but that isn't really the problem here - i think the problem is that we look for a `.vim` at all

does anyone know why we would do that instead of looking for `$HOMEDIR/.vim` or some such thing?

**alternatively**, we could try/catch the read instead of bailing (so we'd resolve to the windows settings after all)

## Reproduce the bug

- Install vim on windows
- Install vim on ubuntu via WSL
- Install coc to both
- Open _windows_ vim and edit a file within ubuntu (importantly, also within your homedir)
- See that it resolves to the linux `coc-settings.json` (and arguably shouldn't since windows has its own)

43081j:
Currently, we will try to read configuration from disk and leave any exceptions unhandled. This can result in unexpected bailing when the file cannot be read for whatever reason.

This change will ignore such failures, treating them the same as if the path didn't exist.

If you'd rather do this another way, let me know and i'll update

Fixes #5117

### DIFF
--- a/src/configuration/index.ts
+++ b/src/configuration/index.ts
@@ -188,10 +188,10 @@ export default class Configurations {
     return undefined
   }
 
-  public parseConfigurationModel(filepath: string | undefined): ConfigurationModel {
+  public parseConfigurationModel(filepath: string | undefined, filecontents?: string): ConfigurationModel {
     if (!filepath || !fs.existsSync(filepath)) return new ConfigurationModel()
     let parser = new ConfigurationModelParser(filepath)
-    let content = fs.readFileSync(filepath, 'utf8')
+    let content = filecontents || fs.readFileSync(filepath, 'utf8')
     let uri = URI.file(filepath).toString()
     parser.parse(content)
     if (!isFalsyOrEmpty(parser.errors)) {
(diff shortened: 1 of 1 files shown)


fannheyward (approved):

fannheyward:
@all-contributors please add @43081j for code.

PR #5132, opened 2024-08-31
fix(typings): getDocument return type.

weirongxu:


fannheyward (approved):

weirongxu:
Do we need to keep this undefined result or modify unit test to migrate to null?

weirongxu:
Many unit tests assert that getDocument is undefined.

https://github.com/neoclide/coc.nvim/actions/runs/10643037745/job/29506143934#step:9:59

weirongxu:
To prevent incompatibility, I think we should also declare both null and undefined.

fannheyward (approved):

Issue #5134, opened 2024-09-02
Some characters cannot be deleted with <C-w> in CocList


statiolake:
## Result from CocInfo

```
## versions

vim version: NVIM v0.10.1
node version: v18.17.1
coc.nvim version: 0.0.82-aabfc231 2024-08-26 07:41:08 +0100
coc.nvim directory: /home/(user)/.local/share/nvim/lazy/coc.nvim
term: tmux
platform: linux
```

## Describe the bug

Some characters (CJK characters and whitespaces) cannot be deleted with `<C-w>` in CocList.

## Reproduce the bug

**We will close your issue when you don't provide minimal vimrc and we can't
reproduce it**

- Open some list by e.g. `:CocList files`
- Press `<Space>` key (or put CJK characters like `あ`)
- Pressing `<C-w>` does not delete the character. `<BS>` normally works.

fannheyward:
https://github.com/neoclide/coc.nvim/blob/b3bcdcbfb3448e998305f5e3c97425e2f9816d88/src/list/prompt.ts#L170

Maybe this regex related.

fannheyward:
@statiolake please give this regex a try

```
pre.replace(/[\w$\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}]+([^\w$\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}]+)?$/u, '')
```

statiolake:
@fannheyward Thanks! Sorry for the late reply.

I tried current master branch and confirmed that it actually removes CJK characters. But now the entire text including the whitespace is removed, e.g. `Hello world<C-w>` removes entire `Hello world`.

I tried the regex of your last comment and it works just fine.

> @statiolake please give this regex a try
> 
> ```
> pre.replace(/[\w$\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}]+([^\w$\p{Script=Han}\p{Script=Hiragana}\p{Script=Katakana}]+)?$/u, '')
> ```

It seems this is because the final commit contains extra `\s` compared to your last comment. Are there any reason to have `\s`?

https://github.com/neoclide/coc.nvim/blob/640544d52afa211fd162590b1bbc7d65bed2813a/src/list/prompt.ts#L170

fannheyward:
> Press <Space> key

As in your step I added `\s` to include the space.

Can you make a PR to remove it?

statiolake:
[...]

PR #5144, opened 2024-09-20
Implement insert-mode like word deletion logic in coc lists

statiolake:
This is some improvement for #5134.

Sorry for bothering you again. Unfortunately, the previous improvement #5138 does not fix the problem correctly - previous implementation still couldn't remove some ending symbols like '-'. This update introduces more refined word detection logic, similar to Neovim's insert mode behavior.

In addition to the fix of above issue, word boundaries are now detected between characters of different classes, just as in the insert mode. For example, in a Japanese sentence, <C-w> works as follows:

今日はいい天気です
今日はいい天気
今日はいい
今日

This behavior stems from recognizing '気' as an 'ideograph' and 'で' as 'hiragana' - different character classes. I've adopted Neovim's character classification to achieve similar functionality. Note that I'm not meant to achieve completely Neovim-compatible behavior.

Moreover, this implementation ensures at least one character is always removed. Consequently, the issue of no characters being deleted, as observed in previous implementations, should no longer occur.

### DIFF
--- a/src/list/prompt.ts
+++ b/src/list/prompt.ts
@@ -1,8 +1,9 @@
 'use strict'
 import { Neovim } from '@chemzqm/neovim'
-import { ListMode, ListOptions, Matcher } from './types'
 import { Emitter, Event } from '../util/protocol'
+import { getUnicodeClass } from '../util/string'
 import listConfiguration from './configuration'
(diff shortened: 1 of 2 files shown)


fannheyward (approved):

fannheyward:
@all-contributors please add @statiolake  for code.

PR #5190, opened 2024-11-13
improve performance for large file

zoumi:
improve performance for large file
(still don't know why this func is called for a buf with b:coc_enabled = 0)


fannheyward:
@all-contributors please add @zoumi for code.

PR #5192, opened 2024-11-19
fix(inlayHint&semanticTokens): Fix range send to server

tcx4c70:
From LSP spec [1]:
> .. the end position is exclusive. If you want to specify a range that
> contains a line including the line ending character(s) then use an end
> position denoting the start of the next line.

but it doesn't mention what we should send to server if the file hasn't EOL. VS Code will set end = { line = last line number (0-based), character = length of the last line}. This fix will change the behavior to the same as VS Code so that some language servers (such as roslyn language server) will work normally.

[1] https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range

### DIFF
--- a/src/model/textdocument.ts
+++ b/src/model/textdocument.ts
@@ -60,6 +60,24 @@ export class LinesTextDocument implements TextDocument {
     return this.lines.length + (this.eol ? 1 : 0)
   }
 
+  public intersectWith(range: Range): Range {
+    let start: Position = Position.create(0, 0)
+    if (start.line < range.start.line) {
+      start = range.start
+    } else if (range.start.line === start.line) {
+      start = Position.create(start.line, Math.max(start.character, range.start.character))
+    }
+
+    let end: Position = this.end
+    if (range.end.line < end.line) {
+      end = range.end
+    } else if (range.end.line === end.line) {
+      end = Position.create(end.line, Math.min(end.character, range.end.character))
+    }
+
+    return Range.create(start, end)
+  }
+
   public getText(range?: Range): string {
     if (range) {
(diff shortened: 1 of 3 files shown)


src/model/textdocument.ts
```diff
@@ -60,6 +60,24 @@ export class LinesTextDocument implements TextDocument {
     return this.lines.length + (this.eol ? 1 : 0)
   }
 
+  public intersectWith(range: Range): Range {
+    let start: Position = Position.create(0, 0)
+    if (range.start.line > start.line) {
+      start = range.start
+    } else if (range.start.line === start.line) {
+      start = Position.create(start.line, Math.max(start.line, range.start.character))
```
fannheyward, line 68: `Math.max(start.character, range.start.character)`

fannheyward (changes requested):

src/model/textdocument.ts
```diff
@@ -60,6 +60,24 @@ export class LinesTextDocument implements TextDocument {
...
+    }
+
+    let end: Position = this.end
+    if (range.end.line < end.line) {
+      end = range.end
+    } else if (range.end.line === end.line) {
+      end = Position.create(end.line, Math.min(end.character, range.end.character))
```
fannheyward, line 75: Why `Math.min`? I can't get this.

fannheyward (changes requested):

src/model/textdocument.ts
tcx4c70: Oops, will the typo

fannheyward (approved):

PR #5196, opened 2024-11-21
feat(neovim): bundled neoclide/neovim

### LINKED ISSUE #5194: `nvim.getOption` fails on nvim 0.11+
The `nvim_get_option` function was deprecated, would return error `Invalid option name`.

The https://github.com/neoclide/neovim package needs to upgrade to use `nvim_get_option_value` function.

fannheyward:
Closes #5194

I bundled the package because I couldn't update `@chemzqm/neovim` on npmjs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "coc.nvim-master",
       "version": "0.0.82",
       "dependencies": {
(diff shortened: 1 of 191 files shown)


davidosomething:
is @chemzqm unreachable?

Issue #5017, opened 2024-05-21
LinkedEdition not work in vim


aispeaking:
## versions

vim version: VIM - Vi IMproved 9.0 9001563
node version: v20.12.0
coc.nvim version: 0.0.82-5c3bdc0 2024-05-21 17:33:35 +0800
coc.nvim directory: /home/liyt/code/extern/vim-runtime/plugged/coc.nvim
term: dumb
platform: linux

## Log of coc.nvim

2024-05-21T19:43:14.931 INFO (pid:209402) [extension:coc-git] - Looking for git in: git
2024-05-21T19:43:14.959 WARN (pid:209402) [workspace] - Extension "undefined" registered synchronized autocmd "VimLeavePre", which could be slow.
2024-05-21T19:43:15.230 INFO (pid:209402) [plugin] - coc.nvim initialized with node: v20.12.0 after 480
2024-05-21T19:43:15.306 INFO (pid:209402) [services] - LanguageClient XML LSP state change: stopped => starting
2024-05-21T19:43:15.309 INFO (pid:209402) [language-client-index] - Language server "xml" started with 209454
2024-05-21T19:43:15.591 INFO (pid:209402) [services] - LanguageClient XML LSP state change: starting => running
2024-05-21T19:43:15.594 INFO (pid:209402) [services] - service xml started
2024-05-21T19:43:15.694 INFO (pid:209402) [attach] - receive notification: doAutocmd [ 6 ]
2024-05-21T19:43:17.503 WARN (pid:209402) [workspace] - workspace.showMessage is deprecated, please use window.showMessage instead.
    at Source.doComplete (/home/liyt/.config/coc/extensions/node_modules/coc-ultisnips/index.js:13:35)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
2024-05-21T19:43:17.505 ERROR (pid:209402) [provider-manager] - Provider error on provideSelectionRanges: Canceled: Canceled
    at LanguageClient.handleFailedRequest (/home/liyt/code/extern/vim-runtime/plugged/coc.nvim/build/index.js:74689:21)
    at /home/liyt/code/extern/vim-runtime/plugged/coc.nvim/build/index.js:69389:31
    at runNextTicks (node:internal/process/task_queues:60:5)
    at process.processImmediate (node:internal/timers:449:9)
    at async Promise.allSettled (index 0)
[...]

PR #5282, opened 2025-03-15
Improve snippets

chemzqm:
- [x] Insert nested snippets instead of merge snippet.
- [x] Different python snippet use its own global values as expected.
- [x] Accept all acceptable changes (not only the change for one placeholder).
- [x] Avoid execute python when current session have snippet using python.
- [x] Use `<Cmd>` for jump key mappings.
- [x] Add method `snippetManager.insertBufferSnippet`


fannheyward:
SnippetTextEdit is added in LSP 3.18, we also need to support this https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#snippetTextEdit

chemzqm:
> SnippetTextEdit is added in LSP 3.18, we also need to support this https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#snippetTextEdit

What should happen when multiple snippets added to the same file?

fannheyward:
> What should happen when multiple snippets added to the same file?

AFAIK, only rust-analyzer server supports SnippetTextEdit feature, I'll give it a try.

chemzqm:
@fannheyward Added the method `snippetManager.insertBufferSnippet` for insert snippet to specific buffer.
Still have issue when typing fast.

src/model/document.ts
```diff
@@ -646,7 +646,7 @@ export default class Document {
       let prev = this._applyQueque.shift()
       if (!prev) fireLinesChanged(this.bufnr)
       this._noFetch = false
-      void this._fetchContent()
+      this.fetchContent()
```
fannheyward, line 649: @chemzqm https://github.com/neoclide/coc.nvim/issues/5017

chemzqm:
@fannheyward Should be good enough to merge, can you try this out?

fannheyward:
> @fannheyward Should be good enough to merge, can you try this out?

Will test this improvement tomorrow, it's a very busy starting month of the year, didn't have enough time nearly 😮‍💨

chemzqm:
> > @fannheyward Should be good enough to merge, can you try this out?
> 
> Will test this improvement tomorrow, it's a very busy starting month of the year, didn't have enough time nearly 😮‍💨

Snippet 部分应该没什么问题，你看看其它部分有没有会受到影响的

fannheyward:
@chemzqm 我重点测试了补全、format 功能，目前没有受到影响

PR #5292, opened 2025-03-26
Use listener_add for buffer changes on vim

chemzqm:
No need to fetch all buffer lines from vim on change anymore.

Tested cases:

- [x] Change, add, delete lines of current buffer.
- [x] Undo and redo
- [x] Join and split the lines
- [x] Change lines with nvim's API and `document.applyEdits`
- [ ] Complicated changes with different kind of change at the same time (use redraw command might help)
- [ ] Change of hidden buffers.

**Note** unlike neovim, using vim's API like `nvim.call('setline', ['.', 'foo'])` won't trigger any buffer change event from vim, unless `redraw` or `listener_flush()` is used.


chemzqm:
>  Complicated changes with different kind of change at the same time

@fannheyward I can't figure out how to make it happen, any ideas?

fannheyward:
> how to make it happen

Have no idea about this.

chemzqm:
Replace lines may cause different kind of change at the same time, didn't observer issue.

Issue #5298, opened 2025-04-03
Popup window not displayed correctly when setting `vim.opt.winborder` in nvim


danielrainer:
## versions

vim version: NVIM v0.11.0
node version: v22.14.0
coc.nvim version: 0.0.82-5866164f 2025-04-03 21:20:47 +0800
coc.nvim directory: /home/drainer/projects/coc.nvim
term: xterm-kitty
platform: linux

## Log of coc.nvim

2025-04-03T23:10:01.375 INFO (pid:29281) [extension:coc-git] - Looking for git in: git
2025-04-03T23:10:01.523 INFO (pid:29281) [plugin] - coc.nvim initialized with node: v22.14.0 after 299
2025-04-03T23:10:01.525 INFO (pid:29281) [services] - LanguageClient Sumneko Lua Language Server state change: stopped => starting
2025-04-03T23:10:01.536 INFO (pid:29281) [language-client-index] - Language server "sumneko-lua" started with 29299
2025-04-03T23:10:01.671 INFO (pid:29281) [services] - LanguageClient Sumneko Lua Language Server state change: starting => running
2025-04-03T23:10:01.678 INFO (pid:29281) [services] - service sumneko-lua started
2025-04-03T23:10:09.139 INFO (pid:29281) [attach] - receive notification: doHover []

## Describe the bug

When setting `vim.opt.winborder` to some allowed value other than `none`, e.g. `single`, in nvim 0.11, the area for the scrollbar is always shown, even when none is needed. In addition, the rightmost text column is cut off.

## Reproduce the bug

- `init.lua`:
```lua
vim.opt.rtp:prepend("$HOME/projects/coc.nvim")
vim.opt.winborder = "single"
vim.keymap.set({ "n" }, "K", function() vim.fn.CocActionAsync("doHover") end, { silent = true })
vim.g.coc_global_extensions = {
	"coc-sumneko-lua",
}
```

- `nvim -u init.lua init.lua`
- `:CocInstall`
- Put the cursor on something for which completions can be shown (e.g. a Lua string)  and press <s-k>. Observe the popup window. In the case of a Lua string (like `"coc-sumneko-lua"` in the example below) the popup has a single line but also a scrollbar area, which should not be there, and the rightmost character is cut off. (It should say `15 bytes` but the last column is not displayed, making it `15 byte`.

## Screenshots (optional)

![Image](https://github.com/user-attachments/assets/884fa988-feed-4255-83e7-7f29eaa58e6e)

[...]

PR #5300, opened 2025-04-06
Make use of the new CocFloatBorder highlight group

danielrainer:
This makes use of the `CocFloatBorder` highlight group introduced in d825dbae9e59fb17a64d172b779ea8bc795aecf5. See also #5298. Without these changes, the `CocFloatBorder` highlight group is not used.

I searched the codebase for usage of `CocFloating` and replaced it with `CocFloatBorder` where it looked to me like it would be appropriate. I might have missed some spots and maybe not all changes in the PR should be applied.

### DIFF
--- a/autoload/coc/float.vim
+++ b/autoload/coc/float.vim
@@ -1463,7 +1463,7 @@ function! s:scroll_win(winid, forward, amount) abort
 endfunction
 
 function! s:get_borderhighlight(config) abort
-  let hlgroup = get(a:config, 'highlight',  'CocFloating')
+  let hlgroup = get(a:config, 'highlight', 'CocFloating')
   let borderhighlight = get(a:config, 'borderhighlight', v:null)
   if empty(borderhighlight)
     return hlgroup
--- a/autoload/coc/dialog.vim
+++ b/autoload/coc/dialog.vim
@@ -674,7 +674,7 @@ function! s:create_prompt_win(bufnr, title, default, opts) abort
         \ 'title': a:title,
         \ 'lines': s:is_vim ? v:null : [a:default],
         \ 'highlight': get(a:opts, 'highlight', 'CocFloating'),
-        \ 'borderhighlight': [get(a:opts, 'borderhighlight', 'CocFloating')],
+        \ 'borderhighlight': [get(a:opts, 'borderhighlight', 'CocFloatBorder')],
         \ }))
 endfunction
 
--- a/data/schema.json
+++ b/data/schema.json
@@ -1204,7 +1204,7 @@
       "type": ["string", "null"],
       "default": null,
       "scope": "application",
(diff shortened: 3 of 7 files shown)


autoload/coc/float.vim
```diff
@@ -1463,7 +1463,7 @@ function! s:scroll_win(winid, forward, amount) abort
 endfunction
 
 function! s:get_borderhighlight(config) abort
-  let hlgroup = get(a:config, 'highlight',  'CocFloating')
+  let hlgroup = get(a:config, 'highlight', 'CocFloatBorder')
```
chemzqm, line 1466: Should not change this one, it should return background highlight group of float window, not border.

chemzqm (changes requested):

autoload/coc/float.vim
danielrainer: Ok, I removed this change.

danielrainer: With this change removed, the borders do not use the `CocFloatBorder` highlight group. The borders are only highlighted using `CocFloatBorder` when explicitly setting `"floatFactory.floatConfig.borderhightlight": "CocFloatBorder"` in `coc-settings.json`, even though this should be the default.

chemzqm: Yes, it's using composed highlight group to make sure the border and float window have the same background color.

danielrainer: Ok, but then I think that the next line should be modified such that the default value of `get` is the `CocFloatBorder` highlight group instead of `v:null`, so: 

```diff
diff --git a/autoload/coc/float.vim b/autoload/coc/float.vim
index 512545ad..9f98ea73 100644
--- a/autoload/coc/float.vim
+++ b/autoload/coc/float.vim
@@ -1455,7 +1455,7 @@ endfunction
 
 function! s:get_borderhighlight(config) abort
   let hlgroup = get(a:config, 'highlight', 'CocFloating')
-  let borderhighlight = get(a:config, 'borderhighlight', v:null)
+  let borderhighlight = get(a:config, 'borderhighlight', 'CocFloatBorder')
   if empty(borderhighlight)
     return hlgroup
   endif
```

danielrainer: And wouldn't it make sense to respect the background color of the borderhighlight group if it is set?

chemzqm: > highlight group instead of v:null

Yes, could change that

> respect the background color of the borderhighlight group if it is set

I prefer not, lots of user not familiar highlight group, the different background color of border only makes the screen ugly.

danielrainer: > Yes, could change that

I opened #5309 for that.

> I prefer not, lots of user not familiar highlight group, the different background color of border only makes the screen ugly.

Ok, that makes it simpler. Having a different background color for the border compared to the rest of the float might be useful in some cases, but I don't need it for my config at the moment.

PR #5312, opened 2025-04-11
perf(vimL): avoid unnecessary has call

fannheyward:
Inspired by https://github.com/neoclide/coc.nvim/commit/0f0f5cbe48b97275c7aa3b47414dea258b63c6bf


plugin/coc.vim
```diff
@@ -508,10 +509,10 @@ function! s:Highlight() abort
     hi default link CocFloatThumb        PmenuThumb
     hi default link CocFloatSbar         PmenuSbar
   endif
-  if has('nvim') && hlexists('FloatBorder')
-    hi default link CocFloatBorder FloatBorder
-  else
+  if s:is_vim
```
chemzqm, line 512: It's not correct, the highlight group would not be defined on nvim without FloatBorder defined.

chemzqm (changes requested):

plugin/coc.vim
fannheyward: @chemzqm got your point, we only defined `CocFloatBorder` on `FloatBorder` exists, other wise fallback to `CocFloating` on vim & nvim both.

PR #5325, opened 2025-04-29
Fix navigate on vim9 can be slow

chemzqm:
When change current complete item on vim9, vim could be really laggy, can be reproduced by start completion before 
https://github.com/neoclide/coc.nvim/blob/master/src/plugin.ts#L53 and hold on the key to trigger `coc#pum#next`.

The solution is use `set completeopt=noinsert,noselect` instead of `set completeopt=menu` to replace text before cursor,
 it's many times faster, but we have to feed `<C-n>`, and `call coc#rpc#request('CompleteStop', [kind])` need to be wrapped with `timer_start` to wait for vim to send lines and TextChangeI event to NodeJS.

The break change is functions like `coc#pum#stop()` `coc#pum#cancel()` `coc#pum#confirm()` `coc#pum#select()` would not have completion done status like before (because of the timer), but they should mostly used by key-mappings, should not break things unless using functions like `getline` just after the function get called.


fannheyward:
Dot trigger is broken with this PR, input `this.`, the suggestions list didn't refresh, reproduced with both nvim and vim9.

```
set nocompatible
set runtimepath^=~/.local/share/nvim/plugged/coc.nvim
set hidden
filetype plugin indent on
syntax on
```

chemzqm:
> Dot trigger is broken with this PR, input this.

Have you build javascript by `node esbuild.js`?

fannheyward:
> Have you build javascript by `node esbuild.js`?

Yes

fannheyward:
The list didn't refresh after the `.`, continue inputting with two more characters will do.

chemzqm:
What's the verbose log? Use `let $NVIM_COC_LOG_LEVEL = 'trace'` in vimrc and open the log by `:CocOpenLog`

fannheyward:
Sorry I'm on the train now, no computer on hands to debug, will test it at night.

chemzqm:
No worry, I guess its issue with your buffer's `iskeyword` option.

chemzqm:
Find the issue, something wrong with the commit character feature.

PR #5329, opened 2025-05-03
Guard against unlisted or invalid buffers

jerryyin:
When attempting to open new directory as buffer, I'd bump into the below error message:

> Error detected while processing function <SNR>1_opendir[4]..BufEnter Autocommands for "*"..function <SNR>82_HandleBufEnter:
line    3:
E158: Invalid buffer name: 25

My interpretation of the error is that something on the Coc side is trying to interpret the buffer number (25) as a buffer name (a path string), and fails when the buffer is invalid (e.g., it's a directory or not yet fully loaded). Adding the conditions to guard against unlisted or invalid buffers.

### DIFF
--- a/plugin/coc.vim
+++ b/plugin/coc.vim
@@ -264,12 +264,11 @@ function! s:Autocmd(...) abort
 endfunction
 
 function! s:HandleBufEnter(bufnr) abort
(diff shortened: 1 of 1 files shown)


chemzqm:
Thanks, it should only guard `listener_flush`

PR #5379, opened 2025-07-17
fix(jump): fix the events for `tab drop`

distroy:
When Vim's `tab drop` command is executed, it iterates through all tabs and triggers the `TabEnter` event for each tab. This behavior interferes with scripts that save tab history. To resolve this issue, we implemented a unified solution that fixes the problem.

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -221,13 +221,13 @@ function! s:safer_open(cmd, file) abort
       endif
       let saved = &wildignore
       set wildignore=
-      let l:page_index = tabpagenr()
+      let l:current_page_num = tabpagenr()
       let l:max_page_num = tabpagenr('$')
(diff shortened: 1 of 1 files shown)


fannheyward (approved):

autoload/coc/util.vim
```diff
@@ -221,7 +221,15 @@ function! s:safer_open(cmd, file) abort
       endif
       let saved = &wildignore
       set wildignore=
-      execute a:cmd.' '.fnameescape(a:file)
+      let l:page_index = tabpagenr()
```
fannheyward, line 224:
```suggestion
      let l:current_page_num = tabpagenr()
```

For better naming consistency

fannheyward (changes requested):

fannheyward (approved):

fannheyward (approved):

PR #5380, opened 2025-07-18
feat(extension): support extensionDependencies

fannheyward:
extension can set `extensionDependencies` in package.json to declare dependencies on other extensions. coc.nvim will install the dependencies, and activate dependencies first before active the main extension.

cc @statiolake, you might need this, please give it a try

### DIFF
--- a/history.md
+++ b/history.md
@@ -2,6 +2,10 @@
 
 Notable changes of coc.nvim:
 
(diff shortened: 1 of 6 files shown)


statiolake:
@fannheyward Thank you very much! This works as I expected!

PR #5382, opened 2025-07-18
fix(jump): improve the events for `tab drop`

distroy:
Simulated the events triggered by tab drop when jumping to the target page:
- Added events for leaving the tab.
- Added buffer-related events, as some plugins will depend on these events.
- Fix the problem of no events when navigating within one tab

### DIFF
--- a/autoload/coc/util.vim
+++ b/autoload/coc/util.vim
@@ -222,9 +222,9 @@ function! s:safer_open(cmd, file) abort
       let saved = &wildignore
       set wildignore=
       let l:old_page_idx = tabpagenr()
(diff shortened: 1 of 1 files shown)


fannheyward:
I am a bit confused, does it mean that Tab and Buf events will not trigger automatically, so all events need to be triggered manually?

distroy:
> I am a bit confused, does it mean that Tab and Buf events will not trigger automatically, so all events need to be triggered manually?

Yes, that's correct. Therefore, the entire process of tab/buf creation, leaving, and entering has been fully simulated here.

Issue #5449, opened 2025-10-16
InlayHints after completion confirmed


A4-Tacks:
## Result from CocInfo

```
## versions

vim version: VIM - Vi IMproved 9.1 9011734
node version: v23.9.0
coc.nvim version: 0.0.82-5552eb7 2025-09-20 11:57:09 +0800
coc.nvim directory: /home/lrne/.vim/plugged/coc.nvim
term: dumb
platform: linux

## Log of coc.nvim

2025-10-16T18:23:16.815 INFO (pid:31812) [plugin] - coc.nvim initialized with node: v23.9.0 after 472
2025-10-16T18:23:16.882 INFO (pid:31812) [services] - service tsserver started
2025-10-16T18:23:33.389 INFO (pid:31812) [attach] - receive notification: startCompletion [ {} ]
2025-10-16T18:23:34.296 INFO (pid:31812) [attach] - Request action: stopCompletion [ 'confirm' ]
2025-10-16T18:23:44.457 INFO (pid:31812) [attach] - receive notification: showInfo []
```

## Describe the bug

After confirming completion, the inlayHint of the parameter is located after the completion text

## Reproduce the bug

**We will close your issue when you don't provide minimal vimrc and we can't
reproduce it**

- Create file `mini.vim` with：

  ```vim
  colorscheme habamax

  set nu
  set laststatus=2
  set statusline=%<%f\ %h%m%r%=%-14.(%l,%c%V%)\ %P
  set statusline^=%(%1*[%{coc#status()}%{get(b:,'coc_current_function','')}]%*\ %)

  call plug#begin('~/.vim/plugged')
  Plug 'neoclide/coc.nvim'
  call plug#end()

  inoremap <silent><expr> <cr>
              \   coc#pum#visible() ? coc#pum#confirm()
              \ : "\<C-g>u\<CR>\<c-r>=coc#on_enter()\<CR>"
  inoremap <silent><expr> <tab> coc#pum#visible()
              \   ? coc#pum#next(0)
              \   : coc#refresh()
  nmap <F3> <Plug>(coc-codeaction-cursor)
  xmap <F3> <Plug>(coc-codeaction-selected)
  nmap <C-k> <Plug>(coc-diagnostic-prev)
  nmap <C-j> <Plug>(coc-diagnostic-next)

  augroup coc_user
    autocmd!
    " Update statusline
    autocmd User CocStatusChange redrawstatus
  augroup end
  ```

- Create file `coc-settings.json` with：

[...]

PR #5643, opened 2026-05-22
fix(inlayHint): clear hints on InsertEnter for Vim

fannheyward:
Vim's textprop has no left-gravity option for inline virtual text (the equivalent of Neovim's `right_gravity = false` on extmarks). When text is inserted at the same column as a zero-width inline prop, Vim always pushes the prop together with the new text, so parameter inlay hints drift to the wrong column after a completion confirm — e.g. `foo(bar: v)` becomes `foo(variablebar: )` instead of `foo(bar: variable)`.

There's no Vim-side knob to fix this today, so on Vim we clear all inlay hints on `InsertEnter` and rely on the existing `InsertLeave` path (changedtick comparison) to re-render once the user leaves insert mode. Neovim behavior is unchanged. The clearing is skipped when `inlayHint.refreshOnInsertMode` is enabled, matching the existing short-circuit.

Refs #5449

### DIFF
--- a/src/handler/inlayHint/buffer.ts
+++ b/src/handler/inlayHint/buffer.ts
@@ -123,6 +123,13 @@ export default class InlayHintBuffer implements SyncItem {
     this._changedtick = this.doc.changedtick
     if (this.config.refreshOnInsertMode) return
     this.cancel()
+    // vim's textprop doesn't support nvim's right_gravity like option, will render hints to
+    // wrong column when text inserted. Clear hints during insert and re-render on leave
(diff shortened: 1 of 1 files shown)


A4-Tacks:
I suggest revert this

1. After clearing inlayhints, it will not be displayed until modifications are made, instead of returning to normal mode
2. This clears all inlayhints. Actually, inlayhints that overlap with editing should be cleared

fannheyward:
@A4-Tacks https://github.com/neoclide/coc.nvim/pull/5644

Issue #5411, opened 2025-08-24
Application completion may not have moved the cursor


A4-Tacks:
## Result from CocInfo

```
## versions

vim version: VIM - Vi IMproved 9.1 9011362
node version: v23.9.0
coc.nvim version: 0.0.82-13d0416 2025-07-31 16:32:31 +0800
coc.nvim directory: /home/lrne/.vim/plugged/coc.nvim
term: dumb
platform: linux

## Log of coc.nvim

2025-08-24T08:50:55.614 INFO (pid:15331) [attach] - receive notification: diagnosticNext []
2025-08-24T08:50:56.398 INFO (pid:15331) [plugin] - coc.nvim initialized with node: v23.9.0 after 2151
2025-08-24T08:50:56.414 INFO (pid:15331) [services] - LanguageClient Rust Analyzer Language Server state change: stopped => starting
2025-08-24T08:50:56.463 INFO (pid:15331) [language-client-index] - Language server "rust-analyzer" started with 15362
2025-08-24T08:50:56.556 INFO (pid:15331) [services] - LanguageClient Rust Analyzer Language Server state change: starting => running
2025-08-24T08:50:56.622 INFO (pid:15331) [services] - service rust-analyzer started
2025-08-24T08:51:03.515 INFO (pid:15331) [attach] - receive notification: highlight []
2025-08-24T08:51:54.812 INFO (pid:15331) [attach] - Request action: stopCompletion [ 'confirm' ]
2025-08-24T08:51:54.935 INFO (pid:15331) [snippets-session] - Content change before snippet, reset snippet position
2025-08-24T08:51:54.942 INFO (pid:15331) [snippets-session] - Jump to final placeholder, cancelling snippet session
2025-08-24T08:51:54.950 INFO (pid:15331) [attach] - receive notification: showSignatureHelp []
2025-08-24T08:51:58.728 INFO (pid:15331) [attach] - Request action: stopCompletion [ 'confirm' ]
2025-08-24T08:51:58.756 INFO (pid:15331) [attach] - receive notification: showSignatureHelp []
2025-08-24T08:52:01.061 INFO (pid:15331) [attach] - Request action: stopCompletion [ 'confirm' ]
2025-08-24T08:52:01.090 INFO (pid:15331) [snippets-session] - Content change after snippet
2025-08-24T08:52:01.129 INFO (pid:15331) [snippets-session] - Before and snippet body changed, cancel snippet session
[...]

PR #5660, opened 2026-06-04
fix(vim): synchronize buffer after undo

fannheyward:
Reading buffer content via getline/getbufline inside the listener_add()
callback corrupts the buffer when the flush processes the undo of an
insert-above operation (e.g. `O` then `u` on a single line). This left
coc's Document.lines permanently out of sync with the real buffer.

Defer content reading out of the listener callback: OnBufferChange now
only records change ranges and schedules a 0ms timer; FlushChanges reads
the updated content and notifies the node side afterwards. A single change
is sent incrementally, while multiple changes trigger a full resync to
avoid line-number drift from deferred incremental updates.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/model/document.ts
+++ b/src/model/document.ts
@@ -247,10 +247,19 @@ export default class Document {
         /**
          * Updated line content, **after** the change is made
          */
-        linedata: string[]
+        linedata: string[],
+        /**
+         * When true, `linedata` is the whole buffer and replaces all lines.
+         * Used for the deferred full resync from `autoload/coc/api.vim` (#5524).
+         */
+        replaceAll?: boolean
       ) => {
         this._changedtick = tick
-        lines = [...lines.slice(0, firstline), ...linedata, ...lines.slice(lastline)]
+        if (replaceAll) {
+          lines = linedata.slice()
+        } else {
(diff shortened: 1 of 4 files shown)


fannheyward:
cc @A4-Tacks @atitcreate

src/model/document.ts
```diff
@@ -247,10 +247,19 @@ export default class Document {
         /**
          * Updated line content, **after** the change is made
          */
-        linedata: string[]
+        linedata: string[],
+        /**
+         * When true, `linedata` is the whole buffer and replaces all lines.
+         * Used for the deferred full resync from `autoload/coc/api.vim` (#5524).
+         */
+        replaceAll?: boolean
```
A4-Tacks, line 255: This field seems to have not been assigned a value

A4-Tacks:
~~After quick testing, this seems to have fixed #5411~~

Oh, it seems that #5411 started working after #5655

src/model/document.ts
fannheyward: https://github.com/neoclide/coc.nvim/pull/5660/changes#diff-f6020d18e947119ddca84d9a5493e810a90384779b142f2c63875c34927fe835R429 sets to true

PR #5690, opened 2026-07-07
fix(pum): position menu on concealed lines

fannheyward:
Use editor-relative coordinates for Vim popups so concealed text does not shift the completion menu. Keep wrapped input anchored to the visible word start and add regression coverage for Vim and Neovim.

Closes #5582

Fixed  by Codex


fannheyward:
@A4-Tacks give it a try

A4-Tacks:
This work

PR #5715, opened 2026-08-03
fix(language-client): make dot matching default

fannheyward:
https://github.com/microsoft/vscode-languageserver-node/pull/1221


chemzqm (changes requested):
Need change history.md for behavior change.

PR #5734, opened 2026-08-14
feat(extension): per-extension VM loader with ESM support and API facade

chemzqm:
Rework extension loading and API access on a per-extension VM isolation model.

### Extension loader (VM-based, no global compiler patching)
- Each extension owns one `vm.Context`, one CommonJS module cache and one ESM module cache ([src/extension/loader.ts](src/extension/loader.ts), [src/extension/esm.ts](src/extension/esm.ts)).
- CommonJS executes through `vm.compileFunction` with an extension-local synchronous `require`; `Module.prototype._compile`, `Module.wrap` and `vm.runInContext` are no longer used for extension execution.
- ESM entries and imports execute through `vm.SourceTextModule`; `coc.nvim`, Node builtins, CommonJS and JSON are bridged with `vm.SyntheticModule`. CJS `require()` of ESM is rejected with `ERR_REQUIRE_ESM`, CJS dynamic `import()` of ESM works, cycles and top-level await are supported.
- coc.nvim now starts Node with `--experimental-vm-modules` automatically (Vim launcher + test runner); startup validation only fails when ESM support is required.
- Reload replaces the runtime completely (fresh context, caches, console state, API facade).

### Per-extension enhanced console
- Independent console with isolated timers/counters/groups per extension ([src/extension/console.ts](src/extension/console.ts)); output keeps `extension:<id>` attribution, `require('console')`/`require('node:console')` return a per-runtime facade while the native `Console` class stays available.

### Per-extension API facade
- `createExtensionApi` ([src/extension/facade.ts](src/extension/facade.ts)) gives every runtime one frozen top-level API object: mutable core singletons (workspace, window, commands, languages, sources, services, extensions, diagnosticManager, listManager, snippetManager, events, mcp) are wrapped, immutable value exports are shared directly, and `nvim` stays a live getter. Mutating one extension's facade cannot affect another extension or coc core.
- Registration callbacks (commands, events, language providers, workspace autocmd/content-provider/keymap handlers) are tagged with the owning extension id: command errors are prefixed with `[extension: <id>]`, event/provider errors include the extension id, and provider `__extensionName` now comes from the registration owner instead of parsing stack traces ([src/util/extensionId.ts](src/util/extensionId.ts)). Extension-owned registrations are disposed with the runtime.

### Test improvements
- Text-synchronization client tests run against an in-process language server instead of spawning a child process per test (removes spawn-related timeouts).
- New suites: facade (identity/isolation/parity/classification/cleanup), ESM loader, per-extension console, command lifecycle, end-to-end callback attribution.

## Verification
- Full suite: 125 files / 3651 tests passing (unit + nvim + vim lanes).
- `tsc --noEmit` and `oxlint` clean.
- Line coverage for the new loader/console/facade files: all ≥ 98% (overall 98.15%).


chemzqm:
Tested all known extensions, activated as expected.

PR #5737, opened 2026-08-20
fix(provider): race aggregation against cancellation

chemzqm:
### Motivation

- Prevent provider aggregation from being kept alive by providers that ignore their cancellation token so cancelled requests can return immediately.

### Description

- Add `raceCancellation` to `Manager` to race provider aggregation against a `CancellationToken` and dispose the listener after resolution (`src/provider/manager.ts`).
- Use `raceCancellation` in `InlineCompletionItemManager.provideInlineCompletionItems` to avoid waiting on unresponsive providers (`src/provider/inlineCompletionItemManager.ts`).
- Update the inline completion test to cover a provider that never settles and assert the API returns immediately after cancellation (`src/__tests__/handler/basic.test.ts`).
- Remove an unused `onUnexpectedError` import as part of the cleanup.

### Testing

- Ran TypeScript typecheck with `tsc -p tsconfig.json --noEmit` which succeeded. 
- Ran linter checks with `oxlint` on the modified files which succeeded. 
- Ran the repository test runner for the updated test with `node scripts/test/cli.mjs src/__tests__/handler/basic.test.ts -t 'provider ignores cancellation'` which passed. 
- Attempted `npx jest` but it could not run in this environment due to network/npm registry restrictions (failed to download dependencies).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a86a78427648325b54be1d76bd3ffc5)

### DIFF
--- a/src/__tests__/handler/basic.test.ts
+++ b/src/__tests__/handler/basic.test.ts
@@ -60,28 +60,17 @@ describe('InlineCompletion', () => {
     assert.strictEqual(res.length, 2)
   })
 
-  it('should return empty when token cancelled', async t => {
+  it('should return immediately when a provider ignores cancellation', async t => {
     let doc = await workspace.document
     let pos = await window.getCursorPosition()
     let context: InlineCompletionContext = { triggerKind: InlineCompletionTriggerKind.Automatic }
-    let cancelled = false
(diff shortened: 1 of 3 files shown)


chemzqm:
Codex misunderstood the intent of the change.

PR #5759, opened 2026-09-17
fix(ui): optimise float right pad span

Ceobe33:
compatible with neovim border type: 'rounded', 'double'..

**Before** 
missing the `>` symbol
image

**After** 
`>`: 👋🏻️ hallo 
Image


chemzqm:
What setting do you use in your Neovim config to give every floating window a border?

Ceobe33:
> What setting do you use in your Neovim config to give every floating window a border?

Use the origin neovim api
`vim.o.winborder = 'rounded'`

and hmmm
some checks seem not pass, could I need gathering more modifies to make a pr?
ah not that mean ig

chemzqm:
> some checks seem not pass, could I need gathering more modifies to make a pr?

Not needed — it's a coverage issue. The real problem is that coc.nvim needs to support this way of configuring borders in Neovim. We need to dig into whether we should just use Neovim's built-in border rather than drawing it ourselves for older versions.

Ceobe33:
> > some checks seem not pass, could I need gathering more modifies to make a pr?
> 
> Not needed — it's a coverage issue. The real problem is that coc.nvim needs to support this way of configuring borders in Neovim. We need to dig into whether we should just use Neovim's built-in border rather than drawing it ourselves for older versions.

Thanks for your time to reply, I'd to like follow this feature. Glad to learn from here.

PR #5766, opened 2026-09-22
feat(watcher): use bundled Parcel watcher backends

chemzqm:
coc.nvim currently depends on an external Watchman installation for filesystem events. Without a working Watchman setup, file watching silently becomes unavailable.

This change bundles the native binaries from `@parcel/watcher` and uses them by default on supported desktop platforms:

- use FSEvents on macOS, inotify on Linux, ReadDirectoryChangesW on Windows, and kqueue on FreeBSD
- select the correct architecture and Linux libc binary at runtime
- preserve the existing `FileSystemWatcher` API, glob filtering, and inferred rename events
- normalize case, Unicode, symlink roots, and Windows namespace paths before coalescing events
- keep an explicitly configured `fileSystemWatch.watchmanPath` as an override
- fall back to Watchman when a compatible native watcher cannot be loaded
- add reproducible binary update/check scripts and smoke-test the bundled binaries on Linux, macOS, and Windows CI

Validation:

- `npm run lint`
- `npm run lint:typecheck`
- `node scripts/test/cli.mjs src/__tests__/core/fileSystemWatcher.test.ts` (38 tests)
- `npm run test:watcher-binary` (darwin-x64, fs-events)


fannheyward:
`@parcel/watcher` will prefer to `watchman` on Linux and Windows if it's installed. We may only need to keep `ParcelWatcher`, no more Watchman client handling.

```
@parcel/watcher has the following watcher backends, listed in priority order:

- FSEvents on macOS
- Watchman if installed
- fts (brute force) on Linux and FreeBSD
- FindFirstFile (brute force) on Windows
```

chemzqm:
> We may only need to keep ParcelWatcher, no more Watchman client handling.

Watchman definitely still needs to stay — at least when the user has it configured, Watchman is still in play — and we'd also need to rework the native side of both to support rename detection.

fannheyward:
> at least when the user has it configured
> 
> rework the native side of both to support rename detection

This is my point: I don't know how many users has configured `watchman`, but won’t involve a large number of users. The both sides handling adds too much rework. I think we can allow this small breaking change: drop watchman, focus on the new watcher.

chemzqm:
> > at least when the user has it configured
> > rework the native side of both to support rename detection
>
> This is my point: I don't know how many users have configured `watchman`, but it won't be a large number of users. Handling both sides adds too much rework. I think we can allow this small breaking change: drop watchman, focus on the new watcher.

I put together a compatibility layer — supporting watchman really isn't that much added complexity.

